PR #402, opened 2014-12-08
Added bzip2 formula

bwsblake:
Implementation of bzip2 formula including tests. Passes audit and works on Mac OSX 10.9.4 with latest brew.


bwsblake:
@DomT4 @jacknagel All items addressed.  Sorry that took so long.  I got busy with other projects.

PR #408, opened 2014-12-27
Adding patch to allow for 500k baud rate with screen.

JamesHagerman:
This patch is left over from a patch submitted to GNU Screen screen-devel list back in 2011. Somehow it never made it into screen... but I can't find information on why it hasn't been included. This is the original mailing list post:

https://lists.gnu.org/archive/html/screen-devel/2011-11/msg00012.html

There are a few situations where a 500k baud rate is pretty useful. If nothing else, a fairly large number of modern car's OBDII ports run at 500k baud. Without this patch, screen is unable to keep up with the OBDII port in logging situations.

### DIFF
--- a/screen.rb
+++ b/screen.rb
@@ -17,8 +17,8 @@ class Screen < Formula
     
     # This patch is to enable 500kbps as a possible baud rate
     patch :p3 do
-      url "https://gist.githubusercontent.com/JamesHagerman/4f13173589c3aa2b0ef7/raw/36ab5f55f59f11d01cd5b95547f5cc06ee5c5e08/gistfile1.txt"
-      sha1 "1b643a99adb8bc3881bbec333d77b11772901bef"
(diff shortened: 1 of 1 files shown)


JamesHagerman:
Sure, I'll look into it.

From what I remember from reading through the threads and documentation related to this file, the Screen developers were leaving this section of tty.sh (which compiles to tty.c) open to the "end user" so that they could implement their own baud rates if needed. Really, in this case, "end user" means "the person that compiles this code themselves". That's GNU for ya...

But some of the world has moved on to marginally faster baud rates and the Screen developers haven't kept up with those changes. This is probably because it's a rather slippery slope to add every baud rate ever to their source code, and that they haven't figured out a better way to customize the baud rate to "non-standard" speeds.

I'll reach out to them directly and see what their point of view is.

mistydemeo:
Have you heard back from the Screen developers?

mistydemeo:
Ping.

apjanke:
Closing due to lack of response.

PR #409, opened 2015-01-01
openssh: bump to 6.7

jbergler:
To fix #397


jacknagel:
Done in 7c8eddc because I needed to remove the revision and I wanted to use the (now updated) MacPorts patches.

PR #410, opened 2015-01-02
Update less to 471

myfreeweb:


exalted:
I see this is reverted with d368ccedf631178791a3d29d7d5d38b4bc7c1f57 and 7999a6293307ddee66a3e0710842b0260325b1fc, apparently because @chdiza [suggested](https://github.com/Homebrew/homebrew-dupes/commit/1ea81b04a4e283e5409ec0c4ba4c5d2cc82f6510#commitcomment-9179680) that it is a "beta software".

According to http://www.greenwoodsoftware.com/less/, less-471 was released for beta testing on 14 Dec 2014. :scream: (release notes: http://www.greenwoodsoftware.com/less/news.471.html)

I am not sure why can't we allow "beta" software in Homebrew/homebrew-dupes (or can we?) The beta version of `less` (that is `less-471`) will map most of the emoji characters from unicode table that are not available in `less-458`. This :mask: , because if you `| less` anything using `--RAW-CONTROL-CHARS` you will get stuff like `<U+129187>` rather than :fireworks: .

[We discussed about this ](https://github.com/as-cii/terminal-emojify/issues/1) separately, so I wanted to give you guys a heads-up, and maybe (please!) consider adding back `less-471`, _unless_ there's a real world issue about it.

Thanks!

/cc: @jacknagel

jacknagel:
It's policy to package stable versions.

This could be added as a `devel` version.

myfreeweb:
What does "beta" even mean if it's been "beta" for several months? It's completely meaningless! Policy, policy... You're becoming the next Debian, with its ancient versions of everything >_<

jacknagel:
Can't please everyone.

PR #415, opened 2015-01-12
groff: add conflicts_with mdocml

afh:
Please note that this change should go along with [homebrew #35790](https://github.com/Homebrew/homebrew/pull/35790).


mistydemeo:
The mandoc that was merged renames the one conflicting manpage, so I don't think we need this after all. Thanks!

PR #421, opened 2015-02-05
bzip2: install man page in the correct directory

sjackman:
Previously installed in share/man/man/man1


sjackman:
Fixed by @jacknagel in 875a918b2bd7ab3c573f0b1e77a262d6cc5409e8

PR #422, opened 2015-02-06
ncurses: Symlink libncurses to libcurses

sjackman:
Symlink key header files from include/ncursesw/ to include/.


mistydemeo:
Do other package managers do this? Judging from `pkg-config` and `ncurses5-config`,  believe this layout is intentional.

sjackman:
Yes, both Ubuntu 14.04 and CentOS 5.10 do. Ubuntu 14.04 installs directly in /usr/lib/x86_64-linux-gnu and /usr/include. CentOS 5.10 installs in /usr/lib64 and /usr/include/ncurses and symlinks the header files into /usr/include.

### CentOS 5.10
```
$ cd /usr/lib64; ls -l libcurses.a libcurses.so
lrwxrwxrwx 1 root root 12 Dec 20  2010 libcurses.a -> libncurses.a
lrwxrwxrwx 1 root root 17 Dec 20  2010 libcurses.so -> libncurses.so.5.5
$ cd /usr/include; ls -l curses.h ncurses.h term.h termcap.h
lrwxrwxrwx 1 root root   16 Dec 20  2010 curses.h -> ncurses/curses.h
lrwxrwxrwx 1 root root   17 Dec 20  2010 ncurses.h -> ncurses/ncurses.h
-rw-r--r-- 1 root root 1841 Jan  6  2007 termcap.h
lrwxrwxrwx 1 root root   14 Dec 20  2010 term.h -> ncurses/term.h
```

### Ubuntu 14.04
```
$ cd /usr/lib/x86_64-linux-gnu; ls -l libcurses.a libcurses.so
lrwxrwxrwx 1 root root 12 Mar 22  2014 libcurses.a -> libncurses.a
lrwxrwxrwx 1 root root 13 Mar 22  2014 libcurses.so -> libncurses.so
$ cd /usr/include; ls -l curses.h ncurses.h term.h termcap.h
-rw-r--r-- 1 root root 77809 Mar 22  2014 curses.h
lrwxrwxrwx 1 root root     8 Mar 22  2014 ncurses.h -> curses.h
-rw-r--r-- 1 root root 39733 Mar 22  2014 term.h
-rw-r--r-- 1 root root  3470 Mar 22  2014 termcap.h
```

sjackman:
@mistydemeo Any further thoughts on this?

sjackman:
@jacknagel @mistydemeo The header files of ncurses are usually expected to be found in `$prefix/include` rather than under a `$prefix/include/ncursesw` subdirectory. On Mac OS, for example:

```
~ ❯❯❯ ls /usr/include/{curses.h,form.h,ncurses.h,term.h,termcap.h}
/usr/include/curses.h  /usr/include/ncurses.h /usr/include/termcap.h
/usr/include/form.h    /usr/include/term.h
~ ❯❯❯ ls /usr/include/ncursesw
ls: /usr/include/ncursesw: No such file or directory
```

This PR symlinks key header files into `include` to mirror this directory structure.

mistydemeo:
@BrewTestBot test this please

ncurses.rb
```diff
@@ -55,11 +55,21 @@ def make_libncurses_symlinks
       end
 
       ln_s "libncurses++w.a", "libncurses++.a"
+      ln_s "libncurses.a", "libcurses.a"
+      if OS.mac?
+        ln_s "libncurses.dylib", "libcurses.dylib"
+      else
+        ln_s "libncurses.so", "libcurses.so"
```
mistydemeo: Looking at Ubuntu as an example - it seems that there is no `libcurses` symlink made? Did you find this was necessary?

sjackman: Yes, Ubuntu does create a `libcurses` symlink. See the [comment above](https://github.com/Homebrew/homebrew-dupes/pull/422#issuecomment-73303603). Yes, I did find it necessary. One example is `homebrew/science/samtools`, but there's almost certainly others.

sjackman:
@mistydemeo Any thoughts on merging this patch?

sjackman:
@mistydemeo How are you feeling about merging this patch?

mistydemeo:
Yes, sorry for the delay. Bottles seem to have expired, so I'll rebuild and merge.

mistydemeo:
@BrewTestBot test this please

sjackman:
Thanks, Misty. ncurses does not build with GCC 5 without a patch, so expect another PR today.

PR #430, opened 2015-02-19
make: update style

sjackman:


make.rb
```diff
@@ -11,11 +9,8 @@ class Make < Formula
   depends_on "guile" => :optional
 
   def install
-    args = %W[
-      --disable-dependency-tracking
-      --prefix=#{prefix}
-    ]
-
+    args = "--disable-dependency-tracking",
+      "--prefix=#{prefix}"
```
jacknagel: This change is unnecessary, and also not a style that we use anywhere.

jacknagel:
I would prefer not to take style-only pull requests.

make.rb
sjackman: @mikemcquaid dislikes `%W`.

sjackman:
The reason for the PR was to induce a bottling of `make`. I should have said so in the comment.

make.rb
mikemcquaid: I'd changed my mind fyi @sjackman; we're going by `brew audit --strict` now.

sjackman: I reverted this chunk.

sjackman:
What's the best way to request that a formula be bottled?

mikemcquaid:
Just ask on the issue.

mikemcquaid:
(or PR)

sjackman:
Please bottle make. Thanks!

mikemcquaid:
Added in 1e3ab98.

sjackman:
Thanks, Mike.

PR #432, opened 2015-03-03
Updated HEAD installation method

deiga:
The `--HEAD` installation failed, because the new `tidy-html5` version uses `cmake`


jacknagel:
looks like this was fixed by 0d06340

PR #436, opened 2015-03-08
Add patch 2.7.5

chdiza:
As a side note, `audit` seems to be broken.  Regarding this new formula, it keeps saying that gnu urls should be "ftpmirror.gnu.org".  So I changed it so that both the `url` and `mirror` fields were `ftpmirror.gnu.org`, and `audit` just kept saying that same thing.  I even tried leaving the `mirror` field blank and just using the `url` field with "ftpmirror.gnu.org", and still audit complained.

I finally gave up and made it as you see here.


chdiza:
I think `audit` is forgetting that even a "ftpmirror.gnu.org" url should start with "https".  When I deleted the protocol prefix, `audit` passed.

chdiza:
I see now that the bots are choking on (what I think is) the audit error I mentioned above.

jacknagel:
This already exists as "gpatch" (previously a limitation of the implementation made it impossible for us to name a class "Patch"): https://github.com/Homebrew/homebrew-dupes/blob/master/gpatch.rb

jacknagel:
For future reference, the URLs for GNU stuff should be

```rb
url "http://ftpmirror.gnu.org/..." # no https, it does not work
mirror "https://ftp.gnu.org/gnu/..." # https
```

this will pass audit.

chdiza:
> For future reference, the URLs for GNU stuff should be

Ah, I didn't know that `https://ftpmirror...` wouldn't work.  Perhaps `audit` should have a clearer warning message, one that includes the `http` protocol.  Otherwise, schlubs like me who have been watching the httpsification of URLs in BrewLand will assume `https` is the right thing.

PR #438, opened 2015-03-09
expect: Symlink libexpect5.45.dylib to lib/

sjackman:
`libexpect5.45.dylib` is installed in `lib/expect5.45/libexpect5.45.dylib` and not symlinked into `lib`, which causes trouble for Linuxbrew as per Homebrew/linuxbrew#306.

It seems common to install this library in `lib/`
[Debian](https://packages.debian.org/wheezy/amd64/expect/filelist)
[Fedora](http://rpm.pbone.net/index.php3/stat/4/idpl/27691851/dir/fedora_19/com/expect-5.45-8.fc19.x86_64.rpm.html)
[Linux From Scratch](http://www.linuxfromscratch.org/blfs/view/svn/general/expect.html)


sjackman:
Thanks, Jack.

PR #439, opened 2015-03-10
Replace file-formula with file

jacknagel:


jacknagel:
The fix was in Homebrew itself, but it looks like it's not quite complete enough to re-use this class name yet.

PR #444, opened 2015-03-25
nano 2.4

agordon:
The patch for devel-2.3.6 has been fixed in 2.4 - so removed.

A minor patch is required to compile v2.4 on Mac OS X.
see https://savannah.gnu.org/bugs/?44609 .

### DIFF
--- a/nano.rb
+++ b/nano.rb
@@ -11,6 +11,7 @@ class Nano < Formula
   depends_on "pkg-config" => :build
   depends_on "gettext"
   depends_on "homebrew/dupes/ncurses"
+  depends_on "glib"
 
   def install
     system "./configure", "--disable-debug",
--- a/nano.rb
+++ b/nano.rb
@@ -2,18 +2,11 @@ require 'formula'
 
 class Nano < Formula
   homepage 'http://www.nano-editor.org/'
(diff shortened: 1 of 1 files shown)


mistydemeo:
Looks like it fails on 10.9 due to not finding sprintf or glib:

```
checking for pkg-config... /usr/local/opt/pkg-config/bin/pkg-config
checking for GLIB - version >= 2.0.0... Package glib-2.0 was not found in the pkg-config search path.
Perhaps you should add the directory containing `glib-2.0.pc'
to the PKG_CONFIG_PATH environment variable
No package 'glib-2.0' found
no
*** Could not run GLIB test program, checking why...
*** The test program failed to compile or link. See the file config.log for the
*** exact error that occured. This usually means GLIB is incorrectly installed.
configure: error: 
*** snprintf() not found.  GLIB 2.x not found either.
*** You should install the GLIB 2.x library which can
*** be found at http://ftp.gtk.org/.
```

agordon:
Strange.. it seems to be build fine on the other systems.
I've added 'glib' dependency.
May I ask to test again? (if it works, I'll send a new clean patch).

Thanks!

agordon:
closed (replaced by #445)

mistydemeo:
Thanks!

For reference, if you squash your commits and force-push back to the same branch on Github, the PR will be updated.

PR #446, opened 2015-03-30
instead of keg-only: use program-prefix g by default and introduce --wit...

halloleo:
instead of keg-only: use program-prefix g by default and introduce --with-default-names to overide this


jacknagel:
@BrewTestBot test this please

PR #449, opened 2015-04-06
openssh 6.8p1

alex:


alex:
I realize there's some tickets about dropping OpenSSH entirely. Without wanting to comment on those substantively, I didn't see any rejected PRs for upgrading it, so I assume the policy is that so long as it's here it'll get updated.

PR #462, opened 2015-05-27
Add option to build openssh with libressl

leonklingele:
This PR adds a new build flag `with-libressl` to build openssh with libressl instead of openssl.

For details about libressl visit: http://www.libressl.org/


mistydemeo:
Do you mind organizing this the way curl does it instead?

leonklingele:
Sure. That indeed seems to be the better way.

leonklingele:
PR has been updated.

PR #463, opened 2015-05-27
rsync 3.1.1: Fix for Linuxbrew

sjackman:
@mikemcquaid @adamv Most formula in Homebrew/dupes build fine without any modifications for Linux, and I'd rather not have to maintain a fork of Homebrew/dupes. Would you consider merging this patch?

@mistydemeo What do you for the Homebrew taps for Tigerbrew?


mikemcquaid:
@sjackman I think it'd be better if Linuxbrew had its own tap/formula for these sort of things. As I've said before I'm pretty unhappy with just add `if OS.mac?`.

sjackman:
`if OS.mac?` is pretty expressive of the true situation in this case, which is that these patches really are specific to Mac OS and make are not useful for any other platform. In an alternate universe where Homebrew and Linuxbrew were a single code base, can you think of a tidier way of dealing with patches that apply to a particular OS? Perhaps
```ruby
patch do
    os :macos
    url "https://trac.macports.org/export/127713/trunk/dports/net/rsync/files/patch-crtimes.diff"
    sha1 "cf07fd4bdf09744f16dfd8f605f5f849a651f641"
end
```
I could see this being a possibly useful syntax for specifying a particular version of Mac OS as well.
```ruby
patch do
    os :mavericks
…
end
```

mistydemeo:
> What do you for the Homebrew taps for Tigerbrew?

Tigerbrew-specific logic is usually handled in the official taps; however there's very little actually required, since most Mac-specific stuff actually still applies to older OSs without modification (and any issues are often shared with Snow Leopard/Lion).

mistydemeo:
That said I personally don't mind including Linuxbrew-specific stuff in the taps if there isn't too much of it, since I know it's enough work maintaining a fork of Homebrew, much less of the taps.

mikemcquaid:
Fair enough. Can you use a single `if` for them all?

sjackman:
Absolutely.

mikemcquaid:
Thanks @sjackman!

For future reference the preferred commit message format for simple version updates is `rsync: mark OS X specific patches.`.

sjackman:
Thanks for the not, Mike.

Issue #428, opened 2015-02-13
Drop openssh


jacknagel:
The openssh formula contains a lot of questionable stuff that I'm uncomfortable maintaining. Rather than strip it down and ship something that isn't useful to people, I suggest we remove it and let the community maintain it elsewhere.

/cc @mistydemeo

jacknagel:
See #413 and #418, I don't want to be responsible for this stuff.

chdiza:
It seems to me that the problems all stem from that keychain patch.  I'd rather keep the formula but ditch the patch entirely; let those who really need it roll their own.

Maybe the launchd patch is behind some of this too, in which case, we should just dump it and let those who need it roll their own.

Not all openssh users need those "fancy" things.

mistydemeo:
@jacknagel I'm with you on this - I would rather not continue to maintain these patches. I wouldn't mind maintaining a more vanilla formula, but I do wonder if it's something users would find useful, given the frustration expressed in the past when these patches were dropped.

jacknagel:
>I'd rather keep the formula but ditch the patch entirely; let those who really need it roll their own.

In my experience, people will just keeping submitting issues and PRs re-adding the things that were removed. I'd like to try and avoid that administrative overhead. On the other hand, people may just re-submit an entire openssh formula, so I don't know if it will help.

chdiza:
Perhaps the formula could contain a "please do not try to add such and such" comment, sort of like in the `vim` formula (line 73) where users are begged not to make PR that hardcodes a ruby path.  Or like how there probably should be such a comment in `mutt.rb` telling people to stop asking for the sidebar patch.

h0lzi:
It would be nice to have openssh still around (Because the apple Provided one is broken).
I don't care about the keychain patch and would drop that.

olbrew:
@h0lzi could you elaborate on what's wrong exactly with Apple's provided one?
After having too much problems with the Homebrew version I'm thinking of returning to the Apple one, sure it will not update as fast, but recently it seems like they have done more regular updates of their command line tools package.

jquagga:
Well, as someone who is guilty of preparing to open an issue to add more features to this (add the HPN patch back), I think that even having the "vanilla" stripped down OpenSSH is useful.  The Apple included version of the ssh client is rather old; it's nice to have something newer and to have a base formula I can add patches to if I desire.

h0lzi:
[...]

PR #477, opened 2015-06-21
Add build option for GSSAPI key exchange support.

poffe:
This PR adds a new build option "--with-gsskex" to build openssh with GSSAPI key exchange support.


jacknagel:
I'm probably not going to add any more "feature patches", and may remove the existing ones, per https://github.com/Homebrew/homebrew-dupes/issues/428. Thanks though!

Issue #413, opened 2015-01-08
OpenSSH with keychain support may expose key passphrase


nbrownus:
It's a bit convoluted but I noticed this after struggling and finally succeeding to get `sshd` running on Yosemite.

Steps to reproduce (only tested with private keys that have a passphrase)

- Install `openssh` with `--with-keychain-support`
- Stop apples `sshd`
- Start brew's `sshd`, however you want
- Log into your local machine `ssh localhost`
- In the same terminal, attempt to log into any other host, localhost works here as well `ssh localhost`
- Notice that no matter how hard you try to type in the password prompt your text always appears in the terminal

chdiza:
Is this the issue that MacPorts attempts to fix with the "pam" patch?

https://trac.macports.org/browser/trunk/dports/net/openssh/Portfile?rev=131258

DomT4:
We won't be [supporting the keychain](https://github.com/Homebrew/homebrew-dupes/pull/482#issuecomment-118994372) patch going forwards.

PR #482, opened 2015-07-01
openssh 6.9p1

dconnolly:


Noctem:
The patch URLs and their hashes should be updated to:
* https://trac.macports.org/export/138238/trunk/dports/net/openssh/files/0002-Apple-keychain-integration-other-changes.patch
  * a707f34c9c639ea5963d5040bee0e543cb87e663c0f525933258f4c0e4290acb
* https://trac.macports.org/export/138238/trunk/dports/net/openssh/files/launchd.patch
  * 012ee24bf0265dedd5bfd2745cf8262c3240a6d70edcd555e5b35f99ed070590

DomT4:
We won't be supporting the `0002-Apple-keychain-integration-other-changes.patch` from this release forwards, and it and the corresponding option should be removed from the formula with this version bump.

We are uncomfortable continually supporting a 1900+ line patch which upstream hasn't signed off on that has the potential to both compromise OpenSSH security and Keychain security. From 10.11 it will also be impossible to edit plists in `/System/*` without disabling rootless, which isn't a configuration we'll be intentionally supporting.

It is also the cause of #413 and wanting to drop OpenSSH completely in #428. OpenSSH can remain, but we won't be supporting the keychain integration from this release.

yveslaroche:
El Capitan comes with a newer OpenSSH version.

```sh
❯ ssh -V
OpenSSH_6.7p1, LibreSSL 2.1.3
```

tduzan:
Can we at least get a documented workaround for this behavior change?  I went from 6.8p1 --with-keychain-support to 6.9p1 today and I am unsure how to go about automating the appropriate behavior.  Even using prezto w/ ZSH that should load keys into the ssh-agent, what ends up happening is that I have to use ssh-add for each shell session (each tab in Terminal.app) to avoid being asked my password every time because it's no longer reading out of my login keychain.

Is there some well-understood methodology for ensuring that on profile load my keys are loaded and either explicitly ask for my password that first time etc?

Noctem:
@tduzan My preferred method is to use gpg-agent as my ssh agent, since I'd be using gpg-agent for gpg anyways and I like using pinentry-mac to unlock access. Just `enable-ssh-support` in your gpg-agent configuration, and add:
```shell
export SSH_AUTH_SOCK=$HOME/.gnupg/S.gpg-agent.ssh
gpgconf --launch gpg-agent
```
to your *.zshrc* (or perhaps use Prezto's module). You can also tweak the `default-cache-ttl-ssh` and `max-cache-ttl-ssh` in gpg-agent to change how often you have to type your password.

Tristor:
That's useful @Noctem.  I just submitted a patch for Prezto (a ZSH shell extension I use) to add OS X support into their 'ssh' module, which will cause it work appropriately given an ssh-agent installed via Homebrew.  It remains to be seen whether it'll be accepted.

pjv:
Is there anyway to brew install the prior version of openSSH (6.8p1) --with-keychain-support so I can get back the prior behavior for the interim until upgrade to El Capitan? Or am I really stuck having to compile openSSH with patches from scratch?

wKovacs64:
@pjv  `brew switch openssh 6.8p1`

pjv:
thanks @wKovacs64. 6.8p1 was gone from my Cellar, but I restored it from a time machine backup and then "switch"ed and my ssh life is back to normal.

brycollins:
This change breaks working functionality.
Apples ssh-agent does not work with PKCS#11 smartcards.
Now there is no -l option for Launchd support

Workarounds:
a) Disable Apples ssh-agent Launchd Agent (org.openssb.ssh-agent.plist)  and run ssh-agent manually.
b) revert to the previous openssh.rb and reinstall 6.8p1

DomT4:
Folks, we're aware it removes functionality. If you're comfortable with the problems involved in a multi-thousand word patch that upstream hasn't audited as far as anyone knows, please feel free to set the formula up in your own tap and support the keychain patch again. From 10.11 though, without disabling rootless you will not be able to modify the necessary `plist` in `/System` and that functionality will cease.

#488 is a genuine issue that needs looking at. Simply stating we've removed functionality isn't an issue for Homebrew itself to handle. We will not be supporting the keychain patch going forwards.

@Noctem's workaround is a nice one.

PR #489, opened 2015-07-09
openssh: add https mirror

vszakats:


DomT4:
Merged, Thanks @vszakats.

PR #499, opened 2015-08-09
ncurses 6.0

dunn:


dunn:
Someday I'll remember to test `brew uses`.

PR #500, opened 2015-08-11
openssh 7.0

alex:


openssh.rb
```diff
@@ -1,10 +1,9 @@
 class Openssh < Formula
   desc "OpenBSD freely-licensed SSH connectivity tools"
   homepage "http://www.openssh.com/"
-  url "http://ftp.openbsd.org/pub/OpenBSD/OpenSSH/portable/openssh-6.9p1.tar.gz"
-  mirror "https://fossies.org/linux/misc/openssh-6.9p1.tar.gz"
```
dunn: Here's a different mirror we can use: https://www.mirrorservice.org/pub/OpenBSD/OpenSSH/portable/openssh-7.0p1.tar.gz

alex: Nice find, added.

DomT4:
Merged in https://github.com/Homebrew/homebrew-dupes/commit/b0ee26dfdd8a570cefdd7adfaa16218af95d1276. Thank you for your contribution @alex!

PR #501, opened 2015-08-11
openssh: version bump 7.0p1

tzudot:
openssh: version bump 7.0p1


tzudot:
Did not check Alex's PR.

PR #508, opened 2015-09-02
openssh: only apply patch for OSX

guoxiao:


xu-cheng:
You should this PR to Linuxbrew instead.

xu-cheng:
Sorry, I mistake to think this repo is `homebrew/homebrew`.

Thanks for the pull request! 🎉 Homebrew depends on contributions from community members like you and we're grateful for your support.

PR #510, opened 2015-09-11
remove awk, add nawk 20121220 (new formula)

rkitover:
This is the newest version available of Brian Kernighan's original nawk
awk interpreter. It is of interest to awk programmers testing their
programs on different versions of awk. It is installed as 'nawk' and the
man page as 'nawk.1' and so does not conflict with anything.

The old formula for nawk 'awk.rb' which instals as 'awk' and conflicts
with gawk is removed by this commit.

The man page is edited to refer to 'nawk' instead of 'awk'.

The github repo promises to collect patches and track new versions, it
uses the "gentoo" branch for this, which is set as head. If the author
adds more non-trivial patches to this branch and tags it I will switch
to using the tarballs from github instead. Currently there are just a
couple of trivial patches for the makefile. The "gentoo" branch of the
github repo is set as head.

The homepage and source url is set to Brian's site.

I previously tried to submit this to homebrew main here:

https://github.com/Homebrew/homebrew/pull/43140#event-407127068

### DIFF
--- a/awk.rb
+++ b/awk.rb
@@ -1,20 +1,58 @@
 class Awk < Formula
+  desc "Brian Kernighan's pattern scanning and processing language"
   homepage "http://www.cs.princeton.edu/~bwk/btl.mirror/"
   url "http://www.cs.princeton.edu/~bwk/btl.mirror/awk.tar.gz"
   version "20121220"
   sha256 "8dc092165c5a4e1449f964286483d06d0dbfba4b0bd003cb5dab30de8f6d9b83"
-
-  conflicts_with "gawk",
-    :because => "both install awk executables."
+  head "https://github.com/junghans/nawk.git", :branch => "gentoo"
 
   def install
-    ENV.O3 # Docs recommend higher optimization
-    # the yacc command the makefile uses results in build failures:
-    # /usr/bin/bison: missing operand after `awkgram.y'
-    # makefile believes -S to use sprintf instead of sprint, but the
-    # -S flag is not supported by `bison -y`
-    system "make", "CC=#{ENV.cc}", "CFLAGS=#{ENV.cflags}", "YACC=yacc -d"
-    bin.install "a.out" => "awk"
-    man1.install "awk.1"
+    # have to run bison first or parallel build fails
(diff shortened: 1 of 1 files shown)


mistydemeo:
Let's leave the formula name alone, and update the existing file with your formula.

rkitover:
That kind of defeats the whole purpose of this. This awk is recognized as "nawk" and is generally installed as "nawk", and most of what makes this formula different is renaming it as such (and making parallel build work.)

There is no particular reason to call it 'awk' as there is no standard awk.

But I'll do this if you want.

manphiz:
Agree with @mistydemeo . Add a paragraph of caveat would be better.

rkitover:
Ok, how's this solution? I modeled it after coreutils and solaris.

rkitover:
@manphiz what do you think about this version?

awk.rb
```diff
@@ -1,20 +1,58 @@
 class Awk < Formula
+  desc "Brian Kernighan's pattern scanning and processing language"
   homepage "http://www.cs.princeton.edu/~bwk/btl.mirror/"
   url "http://www.cs.princeton.edu/~bwk/btl.mirror/awk.tar.gz"
   version "20121220"
   sha256 "8dc092165c5a4e1449f964286483d06d0dbfba4b0bd003cb5dab30de8f6d9b83"
-
-  conflicts_with "gawk",
-    :because => "both install awk executables."
+  head "https://github.com/junghans/nawk.git", :branch => "gentoo"
```
manphiz: Is this the actual upstream of nawk where it is maintained now? Just to make sure.

manphiz:
Small question about head branch. Otherwise looks good to me. Pending ack from @mistydemeo.

rkitover:
I believe this guy is the gentoo maintainer, and he promises to collect patches for nawk, it is the only repo that I found:

https://github.com/junghans/nawk

manphiz:
Adapted and merged in https://github.com/Homebrew/homebrew-dupes/commit/4dfe8686d97b8942d197d4ff4b5957471ea2aaff (messed up with the issue number). Thanks for your contribution, and sorry for the delay

PR #517, opened 2015-10-15
ncurses: build low-level terminfo library

iml:
Provides the low-level libtinfo which some software use rather than the full ncurses library (see https://github.com/Homebrew/homebrew-science/pull/2801). Debian packages this as libtinfo-dev. Should not affect the main library, but could be optional instead.


dunn:
The test is hitting up against the sandbox, but it should probably be replaced with something simpler that just tests the library.

apjanke:
I think the `--with-termlib` change is causing a legit test failure, not just a sandbox issue. When I test locally, the existing formula works, but this PR fails with the following.

```
$ brew test ncurses
Testing homebrew/dupes/ncurses
==> Using the sandbox
==> /usr/local/Cellar/ncurses/6.0_1/bin/tput cols
==> /usr/local/Cellar/ncurses/6.0_1/test/configure --prefix=/private/tmp/ncurses20160406-29257-wlv1mz/test --with-curse
==> make install
Last 15 lines from /Users/janke/Library/Logs/Homebrew/ncurses/03.make:
/usr/bin/clang -Os -w -pipe -march=native -mmacosx-version-min=10.9  -o test_vid_puts ../ncurses20160406-29257-wlv1mz/test_vid_puts.o -L/usr/local/Cellar/ncurses/6.0_1/lib -L/usr/local/lib -F/usr/local/Frameworks -Wl,-headerpad_max_install_names -I. -I/usr/local/Cellar/ncurses/6.0_1/test -I../test -DHAVE_CONFIG_H -F/usr/local/Frameworks  -D_DARWIN_C_SOURCE -I/usr/local/Cellar/ncurses/6.0_1/include -D_XPG5 -Os -w -pipe -march=native -mmacosx-version-min=10.9  `echo "-lform -lmenu -lpanel -lncurses -ltermcap  " | sed -e 's/-lform.*-lpanel[^ ]*//'`
...
/usr/bin/clang -Os -w -pipe -march=native -mmacosx-version-min=10.9  -o ncurses ../ncurses20160406-29257-wlv1mz/ncurses.o -L/usr/local/Cellar/ncurses/6.0_1/lib -L/usr/local/lib -F/usr/local/Frameworks -Wl,-headerpad_max_install_names -I. -I/usr/local/Cellar/ncurses/6.0_1/test -I../test -DHAVE_CONFIG_H -F/usr/local/Frameworks  -D_DARWIN_C_SOURCE -I/usr/local/Cellar/ncurses/6.0_1/include -D_XPG5 -Os -w -pipe -march=native -mmacosx-version-min=10.9  -lform -lmenu -lpanel -lncurses -ltermcap
Undefined symbols for architecture x86_64:
  "_use_tioctl", referenced from:
      _main in ncurses.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make: *** [ncurses] Error 1
```

Let's see if that's the same on Jenkins.

@BrewTestBot test this please.

sjackman:
I couldn't get `--with-termlib` to build on Linuxbrew either. A simple solution that works just fine is to symlink `libncurses.so` to `libtinfo.so`. `-lncurses` implements the ABI of `-ltinfo`.

sjackman:
I've submitted a pull request for the `libtinfo.so` symlink. https://github.com/Homebrew/homebrew-dupes/pull/615

manphiz:
Should this be retried?

sjackman:
I belive this patch is no longer need, since https://github.com/Homebrew/homebrew-dupes/pull/615 was merged, which creates a symlink `libtinfo.so`

PR #522, opened 2015-10-19
openssh: Port Apple's LaunchAgent patch

jryans:
Apple's open source version of openssh contains additional changes to allow
ssh-agent to be used as a LaunchAgent which listens on a socket for connections.

A patch is added which applies these changes for the modern version of openssh
in Homebrew.


felixbuenemann:
You should probably submit the patch to the [Homebrew/formula-patches](https://github.com/Homebrew/formula-patches) repo and include it from there.

tdsmith:
We don't have the expertise to review patches to crypto software and have declined to carry this and similar patches; there's some prior discussion in #482 and linked tickets. Thanks for the patch and sorry for the slow review.

PR #528, opened 2015-11-05
openssh: Remove dead mirror

jgonera:
This link doesn't work anymore.


DomT4:
Missed this PR - Apologies. Killed the mirror when I migrated another dead patch a moment ago. Thanks for the submission!

PR #535, opened 2015-12-15
update args using proper berkeley-db option

mishaaq:
With commit 2e9c4ff the optional dependency option "berkeley-db" was updated to use version 4, but the build option used below wasn't updated accordingly. This commit fixes disabling of HDB/BDB backends and allows to build OpenLDAP with Berkeley DB support.

### DIFF
--- a/openldap.rb
+++ b/openldap.rb
@@ -31,7 +31,7 @@ class Openldap < Formula
       --localstatedir=#{var}
     ]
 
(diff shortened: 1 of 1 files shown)


DomT4:
Great spot!

Merged in https://github.com/Homebrew/homebrew-dupes/commit/ac92c606e090eeb62f678f02c2b2b9858ad1486d. Thank you for your contribution to Homebrew @mishaaq; we appreciate it! :smiley_cat:

PR #541, opened 2015-12-28
ncurses: fix pc files installation

veprbl:
```--enable-pc-files``` configure option was broken for some time in the formula. No pkg-config files were actually installed on OSX. On linuxbrew this may lead to a build error:
```
/usr/bin/install: cannot create regular file `/usr/share/pkgconfig/ncurses++w.pc': Permission denied
/usr/bin/install: cannot create regular file `/usr/share/pkgconfig/ncursesw.pc': Permission denied
/usr/bin/install: cannot create regular file `/usr/share/pkgconfig/panelw.pc': Permission denied
```
Original PR #364 is vague on why do we need pc files, so let's assume we need them. The proposed fix updates the way local ncurses pkg-config directory is set. It also introduces an explicit dependency on pkg-config that is required by ```./configure``` before it even considers installing pc files. The dependency is marked as ```:build``` as it's not really required for ncurses. The packages that locate ncurses with pkg-config will have to also ```:build```-depend on pkg-config.


dunn:
Oh, thanks!  I'm guessing the missing `pkgconfig` dependency is why adding that flag wasn't helping in https://github.com/Homebrew/linuxbrew/issues/554.

dunn:
The test is failing, though; it's trying to write outside the sandbox: http://bot.brew.sh/job/Homebrew%20Dupes%20Pull%20Requests/178/version=yosemite/testReport/junit/brew-test-bot/yosemite/test_homebrew_dupes_ncurses/

veprbl:
Fixed

dunn:
Oops, one last thing. Can you add `revision 1` below the `sha256` line?

veprbl:
Done

dunn:
Thanks so much! Merged in 09d96f4fe3c4bdbd7fc2be5e75468df22b863301.

PR #543, opened 2015-12-31
rsync 3.1.2

ilovezfs:
Update to 3.1.2 and use the new upstream rsync-patches tarball.


rsync.rb
```diff
@@ -1,41 +1,25 @@
...
+  url "https://rsync.samba.org/ftp/rsync/rsync-3.1.2.tar.gz"
+  mirror "https://mirrors.kernel.org/gentoo/distfiles/rsync-3.1.2.tar.gz"
+  sha256 "ecfa62a7fa3c4c18b9eccd8c16eaddee4bd308a76ea50b5c02a5840f09c0a1c2"
 
-  bottle do
-    cellar :any
-    revision 1
-    sha256 "164e7c934b2de1b49b1885e96903386ea808cb552a3c9dd9e585f7b98ae865cd" => :yosemite
-    sha256 "96ee2027bfe92f0b5d3f5812eb8b23a46141134ad8308acda9856a591a9ca807" => :mavericks
-    sha256 "3d8b560ebbb6474976804f229ac4861a18aca29ce8541bcb65807096127a3e5f" => :mountain_lion
```
bfontaine: Please don’t remove the `bottle` block; our bot will update it.

ilovezfs: brew audit should be enhanced to detect that! Thanks for the heads up on the proper procedure.

bfontaine: `brew audit` only audits the current version of a formula; it doesn’t know how it looked before so it can’t detect this. I agree we should add an automatic check for that.

ilovezfs: OK, I've put the bottle block back and re-pushed.

DomT4:
I think we'd be happier retaining the existing method of patching. In-formula patching in the manner you've applied is a little problematic in terms of not being obvious what is happening and why, and can potentially fail silently.

Ideally the patches required should be vendored in https://github.com/Homebrew/patches and then that link used here, with an explanation above each patch as to why it is required, if you wouldn't mind. Cheers!

ilovezfs:
There's plenty of precedent in the existing set of Library/Formula for the way I've done this, so unless those all need to be overhauled (issue link?), I'm not sure why this isn't sufficient. Also, since the patches are now distributed in bulk by upstream, there's no reason to expect silent failures.

DomT4:
The only core usage I can find like this is `fakeroot` which is kind of special because the patch is written into the formula inline rather than being under an `__END__` block or fetched externally. That's not a ton of precedent.

ilovezfs:
Also in llvm33.rb and glib.rb, but ok.

ilovezfs:
@DomT4 This should be added to brew audit if it's disallowed.

ilovezfs:
Since the required method is less efficient and will involve more on-going maintenance than what is proposed here, I'll leave this to someone else who sees the wisdom in a less efficient, more maintenance intensive approach.

mikemcquaid:
> less efficient, more maintenance intensive approach

Which is also, as pointed out, more likely to fail in confusing ways for end-users.

ilovezfs:
@mikemcquaid Since these patches are distributed and signed by upstream, part of their commit, release, build, and lsb packaging, all of which anticipate an in tree "patches" directory, I cannot see how versioning and distributing them separately on a homebrew website makes any sense unless homebrew's model of the world is simply broken.

mikemcquaid:
@ilovezfs When they can be applied with a `patch do` block: I'd agree.

ilovezfs:
@mikemcquaid Is a patch do block possible/permissible within the install method with a file:///$PWD/patches url key or some other magical key for relative paths?

mikemcquaid:
No.

ilovezfs:
@mikemcquaid Is there a subclass of abstract download that is able to re-use the same tarball as other patch do blocks while only selectively extracting a single file?

mikemcquaid:
No, we need direct URLs to individual patches, sorry.

ilovezfs:
@mikemcquaid Would that be an enhancement you'd be interested in having?

mikemcquaid:
@ilovezfs Probably not unless there 3+ projects that use these similarly and have no other way of downloading these patches (e.g. from their version control system).

ilovezfs:
@mikemcquaid Seems like a rather unfortunate waste of the flexibility of these long-standing archive formats.

For instance:

```
tar -t --include '*hfs-compression.diff' -f rsync-patches-3.1.2.tar.gz
```

```
 tar -x --strip-components 2 --include '*hfs-compression.diff' -f rsync-patches-3.1.2.tar.gz
```

mikemcquaid:
Given infinite resources: yes, it'd be nice to support that.

ilovezfs:
@mikemcquaid It depends on https://github.com/Homebrew/homebrew/pull/47614

rsync.rb
```diff
@@ -16,22 +17,26 @@ class Rsync < Formula
...
-    patch do
-      url "https://raw.githubusercontent.com/DomT4/scripts/d47dd65dbae/Homebrew_Resources/MacPorts_Import/rsync/r127713/patch-fileflags.diff"
-      mirror "https://trac.macports.org/export/127713/trunk/dports/net/rsync/files/patch-fileflags.diff"
-      sha256 "b50f0ad6d2c20e561e17b64f07545b1ecfe7d61481a6e5af192abfe21af01e73"
+    patch :p1 do
+      url "https://download.samba.org/pub/rsync/src/rsync-patches-3.1.2.tar.gz"
+      mirror "https://www.mirrorservice.org/sites/rsync.samba.org/rsync-patches-3.1.2.tar.gz"
+      mirror "https://launchpad.net/rsync/main/3.1.2/+download/rsync-patches-3.1.2.tar.gz"
+      sha256 "edeebe9f2532ae291ce43fb86c9d7aaf80ba4edfdad25dce6d42dc33286b2326"
+      file "fileflags.diff"
```
mikemcquaid: Feels pretty repetitive; would be good to make this `files` instead and have a single array.

ilovezfs: I'd started in that direction and ran into this unfortunate beast:
https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/resource.rb#L101-L103

ilovezfs:
@mikemcquaid Changed the name from @file to @extracts and @extract to avoid collision with the preexisting "files" method and allow for a single patch block to handle more than one at time, as you suggested. Also, multiple instances of @extracts and @extract are supported.

ilovezfs:
@mikemcquaid With ``apply`` now merged, is there a way for you to re-trigger these tests?

DomT4:
@BrewTestBot test this please

mikemcquaid:
Pushed, really great work here @ilovezfs.

ilovezfs:
@mikemcquaid Thanks! :)

PR #547, opened 2016-01-10
diffstat: desc added

bfontaine:


DomT4:
Thanks for pulling updated bottles where necessary! :heart:

PR #566, opened 2016-02-08
tcl-tk: don't apply patch for sqlite on Linux machines

maxim-belkin:
In a Linux fork of Homebrew, `MacOS.version` returns `0` and, therefore, simple check `MacOS.version < :something` always true on a Linux machine. This fix makes the patch for sqlite inside the tcl-tk formula work only on Mac machines only.


maxim-belkin:
I guess it is easier to make `MacOS.version` on Linux return `9999` xP

dunn:
@BrewTestBot test this please

DomT4:
@sjackman Was a decision ever made on the status of `dupes` for Linuxbrew?

sjackman:
So far small changes like this one have been accepted into Homebrew/dupes. There are similar patches in `ncurses`, `openssh` and `rsync`.

Maxim, I was able to install `brew install tcl-tk --without-tk` without needing this patch. Is that because I used `--without-tk`, or because I already had `sqlite` installed?
Does `brew install sqlite tcl-tk` fix your issue?
Please report `brew gist-logs tcl-tk` without this patch applied.

DomT4:
> So far small changes like this one have been accepted into Homebrew/dupes.

Aye, I'm not unhappy to accept small changes like this. Was more curious as to whether there had been more thought on integrating `homebrew/dupes` into Linuxbrew core more and blacklisting this tap, given Linuxbrew needs quite a bit from it and they aren't often dupes on that platform.

sjackman:
I've already pulled a select few formula like `zlib` into Linuxbrew/core that are dependencies of the compiler toolchain `glibc` and `gcc`. I'd rather not have to periodically copy over the commits from Homebrew/dupes into Linuxbrew/core, although it could likely be automated. Homebrew-dupes doesn't really change the often anyway.

DomT4:
@BrewTestBot test this please

DomT4:
Merged in https://github.com/Homebrew/homebrew-dupes/commit/5df8b5560eed14096e3ce2bd3865a7c297d5fcdb. Thanks @maxim-belkin! :smiley:

PR #570, opened 2016-02-24
Missing simlink for panel.h to `brew --prefix`/incude

palad1:
panel.h is missing in the include directory. Breaks llvm-lldb (and standard distros do place panel.h at the top of /usr/include, ex: centos)


palad1:
Another file is missing, nvm

PR #571, opened 2016-02-24
ncurses.rb add missing panel.h to include

palad1:
Breaks llvm-lldb compilation, most distros put panel.h at the include root, (centos)

### DIFF
--- a/ncurses.rb
+++ b/ncurses.rb
@@ -77,7 +77,7 @@ class Ncurses < Formula
 
     ln_s [
       "ncursesw/curses.h", "ncursesw/form.h", "ncursesw/ncurses.h",
-      "ncursesw/term.h", "ncursesw/termcap.h"], include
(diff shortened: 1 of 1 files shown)


dunn:
If this is something that all users should receive, could you change `revision 1` to `revision 2` below the `sha256`?

DomT4:
Ping?

manphiz:
@palad1 Ping for @dunn 's question.

apjanke:
This seems to make sense, especially since Apple puts `panel.h` in the root.

```
$ find /usr/include -name panel.h
/usr/include/panel.h
```

Can you give us a more specific example of what breaks with the current configuration? The `llvm --with-lldb` formula in core doesn't depend on this.

apjanke:
Merged in #650 (preserving your author credit).

Thank you for your contribution to Homebrew!

PR #573, opened 2016-02-27
lapack: mv lib64/* to lib/* in linuxbrew.

yesimon:
The liblapack libraries are installed into lib64/ using linuxbrew, which are not linked during the linking step. Since lib64 is a symlink to lib anyways, just move them to lib.


sjackman:
I'll fix this issue in Linuxbrew core.

DomT4:
Closing here then.

PR #574, opened 2016-02-28
openssh: Add libedit to requirements

yesimon:
Since it's compiled with libedit, adding it to dependencies. 

It might be better to make this a config option `--with-libedit`.


DomT4:
I _think_ @sjackman prefers `unless OS.mac?` syntax, but could be wrong.

sjackman:
Yes, and the reasoning is that Mac OS provides `libedit`, so the dependency is not needed on that specific platform.

PR #583, opened 2016-03-21
openssh: Add GSSAPI key exchange support

kek:
Add the GSSAPI key exchange patch from MacPorts.

https://trac.macports.org/browser/trunk/dports/net/openssh/files/openssh-7.2p1-gsskex-all-20141021-mp-20160301.patch

### DIFF
--- a/openssh.rb
+++ b/openssh.rb
@@ -15,11 +15,19 @@ class Openssh < Formula
   # Please don't resubmit the keychain patch option. It will never be accepted.
   # https://github.com/Homebrew/homebrew-dupes/pull/482#issuecomment-118994372
   option "with-libressl", "Build with LibreSSL instead of OpenSSL"
+  option "with-gsskex", "Add GSSAPI key exchange support"
 
   depends_on "openssl" => :recommended
   depends_on "libressl" => :optional
   depends_on "ldns" => :optional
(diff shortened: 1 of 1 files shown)


DomT4:
This doesn't fill me with hope:
>  This patch has been rejected upstream: "None of the OpenSSH developers are
in favour of adding this, and this situation has not changed for several
years.  This is not a slight on Simon's patch, which is of fine quality, but
just that a) we don't trust GSSAPI implementations that much and b) we don't
like adding new KEX since they are pre-auth attack surface.  This one is
particularly scary, since it requires hooks out to typically root-owned
system resources."

Unless OpenSSH themselves take the patch, or Apple apply the patch to their OpenSSH for modern OS X releases we're going to be extremely reluctant to complicate the formula further.

kek:
Fair enough. I since found out that we already tried this once in #477. Sorry about the duplicate. As I understand it, we require this for our network setup to work well with Kerberos authentication, but we can probably maintain our own version, then.

DomT4:
No worries. Yeah, we recommend that if people want custom elements of OpenSSH that OpenSSH or Apple aren't interested in touching for modern releases they use their own tap for OpenSSH. Apologies, but thanks for the submission nonetheless.

PR #590, opened 2016-04-03
Update a few formulas to use https:// where possible

reedloden:
* Update the `apple-gcc42` and `awk` formulas to use https:// for downloading files.
* Update the homepages mentioned in `file-formula` and `lsof` to use https:// as well.


reedloden:
(I'm aware there is checksum validation that happens, but wherever https:// is available, best to use it. :)

apjanke:
The tests are failing because the `lsof` URLs aren't working (#591). Otherwise looks good. Merging despite the failure.

Thank you for your contribution to Homebrew!

PR #593, opened 2016-04-10
fix configure: error: Cannot link with GPM library

windscope:
When trying to install the readline package with Linuxbrew on CentOS, got error:
configure: error: Cannot link with GPM library
After adding "--with-gpm=no", installation was succeeded.


DomT4:
Seems reasonable enough.

Merged in https://github.com/Homebrew/homebrew-dupes/commit/a10d5a50d17218b6d4ee36785afe0289375cfab3. Thank you for your contribution to Homebrew @windscope; we appreciate it! :smiley_cat:

PR #597, opened 2016-04-30
less: fix Using GNU Mirror

fuzzyroddis:
GNU has regional mirrors


manphiz:
Merged and added back original link as fallback. Thanks for your contribution.

PR #602, opened 2016-05-05
libpcap: Depends on bison and flex for Linuxbrew

sjackman:
Fix the following error:
```
checking for bison... bison
configure: WARNING: don't have both flex and bison; reverting to lex/yacc
checking for capable lex... insufficient
configure: error: Your operating system's lex is insufficient to compile libpcap.
```


DomT4:
Should be in block format where there are multiple `if` or `unless` that align. Fixed that in [post](https://github.com/Homebrew/homebrew-dupes/commit/1f35039c6c96ffcc9cb5cbdc12d58e11e3c38898).

Thanks @sjackman!

sjackman:
Thanks, Dominyk!

PR #603, opened 2016-05-10
[nano] Add option to build with NLS

cprecioso:


apjanke:
Let's discuss this. Any reason to not make this the default, instead of an option? Anyone know why NLS was disabled in the first place?

I did my due diligence, and don't see anything in the commit or issue history indicating why NLS was disabled in the first place. Maybe I just don't understand what NLS entails. But it seems to be just locale-aware localization of UI messages. From [this page](http://www.nano-editor.org/dist/v2.2/nano.html)

```
--disable-nls
    Disables Native Language support. This will disable use of the available GNU nano translations. 
```

If NLS is just message localization, I think we should just turn it on by default and not provide an option to turn it off.

cprecioso:
That was my reaction too (given I wanted to install nano to have it in my
language). I don't know if this has more far-reaching implications. I left
it off by default because:

- It was that way originally
- A search on `homebrew/dupes` shows `grep` also has this disabled by
default.
- A search on `homebrew/core` showed various formulas, some of which have
this enabled by default and some of which have disabled. The lack of
consensus made me not want to modify the default options.

I think the main reasons to turn this on by default are:

- Building with `nls` didn't seem to affect build time or space occupied
significantly.
- I guess I'm not the only one who wants to install a dupe because it's in
their language.
- The option is actually on by default in the `nano` `./configure` file,
you have to disable it explicitly.
- I couldn't find a reason for this to be disabled in the first place.

DomT4:
Since it doesn't stop the bottle being `cellar :any` or require any additional dependency I'd just toss it in by default. I'm not sure why it was off in the first place, presuming it builds okay on all platforms, but we're not really saving much of anything by keeping as it was prior to this PR.

apjanke:
There's a bunch of `--disable-nls` flags out there; some optional, some not. Let's do a little digging and see if anyone knows why. I would think disabling NLS was an rare enough use case we wouldn't need to provide it.

Some of these look like they're making it optional in order to avoid a `gettext` dependency. But, from looking at the `ABOUT-NLS` file in `nano`'s source distribution, that shouldn't be necessary frequently, because GNU packages that do NLS with `gettext` will vendor `gettext` in their own distributions.

```
[/usr/local/Library/Taps/homebrew on ⇄ master]
$ grep -ri disable-nls *
homebrew-core/Aliases/mc:    args << "--disable-nls" if build.without? "nls"
homebrew-core/Aliases/pipeviewer:    args << "--disable-nls" if build.without? "gettext"
homebrew-core/Aliases/svn:      --disable-nls
homebrew-core/Formula/cogl.rb:               "--disable-nls",
homebrew-core/Formula/exif.rb:    args << "--disable-nls" if build.without? "gettext"
homebrew-core/Formula/gcc.rb:    args << "--disable-nls" if build.without? "nls"
homebrew-core/Formula/gtk-gnutella.rb:    system "./build.sh", "--prefix=#{prefix}", "--disable-nls"
homebrew-core/Formula/liblwgeom.rb:      "--disable-nls",
homebrew-core/Formula/midnight-commander.rb:    args << "--disable-nls" if build.without? "nls"
```

`git blame` is no use so far, since the `--disable-nls` was here on `nano` when it was first imported from another repo.

cprecioso:
I've updated the PR. Now it just builds with NLS by default and leaves no option to disable it, by this conversation.

apjanke:
I'm tentatively :+1: on this. Wish we knew more specifically why it was off in the first place, though.

Since this is now just removing a build flag, would you mind cleaning this PR up a bit? It currently has two commits and moves some lines code around, but the same effect could be achieved with a one line deletion vs the original code. Could you squash this to a single commit and change it to do a minimal change, just deleting the `--disable-nls` argument? Don't worry about the `*args` versus inline arguments style here.

cprecioso:
Done

apjanke:
Sorry I missed this earlier, but this should have a revision bump, since it's changing the result of the build and the bottles should reflect that. Could you add a `revision 1` (after line 6) (and squash to a single commit again)?

Sorry for the slow response time here.

DomT4:
FWIW, we don't need to bump the formula revision unless we want everyone with existing installations to get the changes. If we're happy for _only_ the new bottles & people who install them to gain the NLS functionality a simple pulling of new bottles will work.

apjanke:
I think we may as well bump it so existing installs get the changes, and make that a general policy. To the extent that this is a useful change, there may be folks out there who might want to use NLS, but they'd have to know it was available and manually do a `brew reinstall` (and maybe clear out the bottle cache). Plus if there are issues with it, then we'd have to ask users who have `nano` installed whether they installed it before or after this change to know which build they have. Seems to me that always doing a formula revision bump upon build changes keeps things simpler.

DomT4:
For `nano` I'd agree, since it's tiny & the bottle isn't fond of a single particular home, but for things with longer compile times with prefix-specific bottles I think that's more inconvenience than convenience.

The bottle revision is incremented every time someone `brew pull`s a new one, so unless someone's executed `brew test-bot nano` _inside_ `HOMEBREW_CACHE` it's unlikely there are issues around hitting "old" bottles for new changes.

cprecioso:
Done

apjanke:
Merged. Thank you for your contribution to Homebrew!

sjackman:
This bottle was previously `cellar :any` and no longer is after this PR. If someone were to have the time, I'd appreciate looking into why the `cellar` status changed, and if it's a false-positive or a true effect.

DomT4:
That happens a lot with NLS 😢. We've been trying to figure out why, but truth be told it's low low low on the list of priorities & we'd love someone both invested in NLS & with the time to do so to dive into the issue & file some PRs to help resolve it.

apjanke:
Looks like the path to the `share/locale` directory under the installation prefix gets baked in to the binary. Since that's not a library linkage, that couldn't get fixed with the bottle relocation tool, right? So I think that's a true positive.

```
$ brew bottle -v nano
==> Determining homebrew/dupes/nano bottle revision...
==> Bottling nano-2.6.0.mavericks.bottle.1.tar.gz...
/usr/bin/touch -h -t 201606170945.07 /usr/local/Cellar/nano/2.6.0/bin/rnano
tar cf /usr/local/nano-2.6.0.mavericks.bottle.1.tar nano/2.6.0
gzip -f nano-bottle.tar
Warning: String '/usr/local/Cellar' still exists in these files:
/usr/local/Cellar/nano/2.6.0/bin/nano
 --> match '/usr/local/Cellar/nano/2.6.0/share/locale' at offset 0x255ba

./nano-2.6.0.mavericks.bottle.1.tar.gz
  bottle do
    revision 1
    sha256 "52a76e1f8ae1bf708d8aec6866cdfec5aa90ae24d932c01bac5420543ffc01bb" => :mavericks
  end
```

Does this affect whether we want to make this the default after all?

sjackman:
Because most Linuxbrew installations are in the user's home directory, I have a strong preference for `cellar :any` bottles. It may be possible to set an environment variable to specify the location of the translation files.

sjackman:
With NLS support enabled, the bottle is no less relocatable than with NLS disabled. NLS won't work if it's disabled (clearly). It will work if it's installed in the default `/usr/local` location without any additional configuration. It won't work if installed somewhere non-default without additional configuration, setting an environment variable to specify the location of the locale files, but that's no worse off than if NLS is disabled. For this reason, I would suggest enabling NLS and teaching the is-this-bottle-relocatable logic to ignore `/usr/local/share/locale`, because it is in practice a false positive.

I'm just going on memory that there is an environment variable to specify the location of the locale files. I don't know its name, and I'm not 100% positive that the variable actually exists. Even if it doesn't, enabling NLS and setting the bottle to `cellar :any` is no worse off than what we have now. NLS works if it's installed in `/usr/local` and doesn't otherwise. Currently, NLS never works.

PR #605, opened 2016-05-20
bc: Depends on bison and flex for Linuxbrew

sjackman:
Fix configure: error: cannot find output from lex; giving up


bc.rb
```diff
@@ -14,6 +14,9 @@ class Bc < Formula
 
   keg_only :provided_by_osx
 
+  depends_on "bison" => :build unless OS.mac?
```
DomT4: These should be in a block 😉 

```ruby
unless OS.mac?
  depends_on "bison" => :build
  depends_on "flex" => :build
end
```

sjackman: Right. I think you've given me that note before.

DomT4:
Merged in https://github.com/Homebrew/homebrew-dupes/commit/c3c0e22e96d463cc6fb62dbcd457effc93b2f011. Thanks for making that change.

sjackman:
Thanks for merging.

PR #607, opened 2016-05-20
less: Depends on ncurses for Linuxbrew

sjackman:
Fix the error:
```
checking for tgoto in -ltinfo... no
checking for initscr in -lxcurses... no
checking for initscr in -lncursesw... no
checking for initscr in -lncurses... no
checking for initscr in -lcurses... no
checking for tgetent in -ltermcap... no
checking for tgetent in -ltermlib... no
checking for library containing regcmp... no
checking for working terminal libraries... Cannot find terminal libraries - configure failed
```


DomT4:
Merged in https://github.com/Homebrew/homebrew-dupes/commit/6a05945c944c89e19fe86b9f09066c8c3235cffd. Thanks Shaun 😃

chdiza:
@DomT4 Shouldn't this be fixed an a Linuxbrew-specific tap?  Are we going to have `unless OS.mac?` in *formulas* now?

DomT4:
> Are we going to have unless OS.mac? in formulas now?

I don't love having it here but it's pretty much acceptable in science, dupes and very very occasionally versions. That's not a new thing.

I mounted an attempt a while back to persuade Shaun to hard-block `dupes` from being tapped and just carry these formulae in Linuxbrew core given most of them aren't dupes at all on Linux, but I can understand logistically why that'd be a chunky burden given Linuxbrew's entire team is two people or something like that.

sjackman:
A tidier fix would be a `depends_on :ncurses` requirement. Is that preferable?

DomT4:
Not particularly, to be honest. That permanently beefs out core code with minimal obvious use case.

chdiza:
"Linuxbrew/dupes".  There's going to have to be separate Linux formula taps anyway, once "Homebrew/brew" becomes some sort of cross-platform gem (or whatever is in the works).

PR #608, opened 2016-05-20
tcpdump: Depends on libpcap for Linuxbrew

sjackman:
Fix the following error:
```
checking for local pcap library... not found
checking for pcap-config... no
checking for main in -lpcap... no
configure: error: see the INSTALL doc for more info
```


tcpdump.rb
```diff
@@ -13,7 +13,7 @@ class Tcpdump < Formula
     sha256 "67abede2a0c328caad072fdd7970580b3ba38e4a79b9eed61f2bd0e47c4b9ffb" => :mountain_lion
   end
 
-  depends_on "homebrew/dupes/libpcap" => :optional
+  depends_on "homebrew/dupes/libpcap" => [(:optional if OS.mac?)].compact
```
DomT4: Let's not overcomplicate this syntax more than necessary 😄.

```ruby
depends_on "homebrew/dupes/libpcap" => :optional if OS.mac?
```
Works fine.

sjackman: That doesn't look to be the same thing to me. It's optional on Mac and required on Linux. How about instead
```ruby
depends_on "homebrew/dupes/libpcap" => (OS.mac? ? :optional : [])
```

DomT4: Oh I see.

Let's avoid ternary operators, they upset Mike because a lot of new contributors don't know how to read them. Use block format instead:

```ruby
if OS.mac?
  depends_on "homebrew/dupes/libpcap" => :optional
else
  depends_on "homebrew/dupes/libpcap"
end
```

sjackman: Done.

DomT4:
Merged in https://github.com/Homebrew/homebrew-dupes/commit/bacf3a73c03c4a89d67df997d9578385adbb37e1. Thanks again Shaun.

sjackman:
Thanks for merging, Dominyk.

PR #609, opened 2016-05-20
units: Depends on readline for Linuxbrew

sjackman:
Fix `units: error while loading shared libraries: libreadline.so.6`


DomT4:
Merged in https://github.com/Homebrew/homebrew-dupes/commit/e9caeb325c1844ea0cf07acac5fb62283615634b. Thanks again @sjackman!

PR #611, opened 2016-05-20
lsof: Fix for Linuxbrew

sjackman:


lsof.rb
```diff
@@ -26,7 +26,7 @@ def install
                 "/usr/include", MacOS.sdk_path.to_s + "/usr/include"
 
       mv "00README", "../README"
-      system "./Configure", "-n", "darwin"
+      system "./Configure", "-n", `uname -s`.chomp.downcase
```
DomT4: Calling `uname -s` directly here makes me a bit sad. Are there many possibilities beyond it throwing up `darwin` or `linux`?

sjackman: On Mac OS and Linux the only two possibilities are `darwin` and `linux`. Other kernels I can't say, but they're not supported anyway.

sjackman: On Linuxbrew we have `OS.name` which is `OS.mac? ? "darwin" : "linux"`. It comes in handy reasonably often.

DomT4: `OS.name` is potentially something that could be added to `homebrew/brew` these days given the intention is for the core to be slowly more portable. It might be worthwhile asking Mike to de-fork Linuxbrew brew/core so they can be searched on GitHub (GH's website bars fork code searching for now).

`OS.mac? ? "darwin" : "linux"` seems reasonable to me, it's simple enough for everyone to understand.

ilovezfs: Since `uname -s` is portable, I'm not sure we need to reinvent the wheel unless there's a DSL for it.

DomT4: I'm just very 😷 on chucking vanilla ``` `` ``` onto the end of `system` calls. It's pretty nasty.

ilovezfs: I'm all for the DSL. Calling uname -s inline is nuts but so is if mac this if linux that :)

DomT4: Yeah, I think given the cross-platform desire in `homebrew/brew` it makes sense for `OS.name` to be a thing. I haven't checked but I'd be surprised if the mechanism for it in Linuxbrew is anything but tiny.

ilovezfs: https://github.com/Linuxbrew/linuxbrew/blob/master/Library/Homebrew/os.rb#L12

We'd probably want something a little fancier, like cached value of uname -s because then it's someone else's hard coded string so we can feel a tad cleaner.

sjackman: The current implementation of `OS.NAME` is essentially:
`/darwin/i === RUBY_PLATFORM ? "darwin" : "linux"`

DomT4: Any feelings on this @MikeMcQuaid?

MikeMcQuaid: I'd vote `uname -s` for now and then do something smarter with `HOMEBREW_SYSTEM` later.

ilovezfs: What's the simplest way to get output from `uname -s` here without resorting to backticks, if we'd like to address @DomT4's concern?

MikeMcQuaid: Let's just use backticks here.

MikeMcQuaid: To me `OS.mac?` is something far less pleasant.

ilovezfs: Yeah I cannot imagine a scenario where `uname -s` blows up where something else is not also very wrong. My question was mainly curiosity what the standard brew way of getting system output in a formula is other than backticks, even if we're not going to use that "other way" here.

MikeMcQuaid: There's nothing wrong with backticks if we're not inserting user or formula modifable text. I even think `shell_output` and `Utils.popen_read` hurt readability for people used to Shell/Bash but not Ruby in the cases where we're not splitting arguments or accepting user text CC @xu-cheng FYI too.

DomT4: Just think it looks pretty grim inline. Where this sort of thing pops up I tend to prefer something like:

```ruby
sys = `uname -s`.chomp.downcase
system "./Configure", "-n", sys
```

The two are close enough so you don't have to go hunting for what on earth `sys` is, but equally you're not mixing and matching execution calls on the same line like we're part of the world's least consistent style grab bag.

MikeMcQuaid: 👍

ilovezfs: To me it looks better just as Shaun has it. p =  f(x) ; q = g(p) is less clear than q = g(f(x)) especially when it all fits neatly on one line :)

sjackman:
I agree with Joe. Creating temporary variables doesn't help readability IMO.

DomT4:
Fed up of discussing the style nit given it's just personal preferences on top of other personal preferences. It's dupes, code consistency tends to go to hell in the taps anyway.

Merged in https://github.com/Homebrew/homebrew-dupes/commit/72879dfdd9989e1238dc1ca44a137124908c9ff7. Thanks folks.

sjackman:
Thanks for merging, Dominyk. Every formula in Homebrew/dupes builds on Linuxbrew now.

PR #613, opened 2016-05-29
unzip: Ubuntu patchset, ZIP64 and symlink support

felixbuenemann:
* Use Ubuntu patchset for unzip, with [lots of bug and security fixes](http://changelogs.ubuntu.com/changelogs/pool/main/u/unzip/unzip_6.0-20ubuntu1/changelog).
* Enable large file support / ZIP64 for >=4GB and >=64k files.
* Compile against system libbz2 instead of built-in.


unzip.rb
```diff
@@ -7,12 +7,1215 @@ class Unzip < Formula
 
   keg_only :provided_by_osx
 
+  # Ubuntu unzip-6.0-20ubuntu1 patchset:
+  # http://changelogs.ubuntu.com/changelogs/pool/main/u/unzip/unzip_6.0-20ubuntu1/changelog
+  patch :DATA
```
DomT4: I'm not touching an 1100+ line patch inline. Just no way 😄. Submit it to https://github.com/Homebrew/formula-patches please.

If it has a security impact the formula should be revisioned, as well.

unzip.rb
```diff
@@ -7,12 +7,1215 @@ class Unzip < Formula
...
+
+  depends_on "bzip2"
```
DomT4: Does this have to be mandatory?

unzip.rb
```diff
@@ -7,12 +7,1215 @@ class Unzip < Formula
...
+
   def install
-    system "make", "-f", "unix/Makefile", "macosx"
+    system "make", "-f", "unix/Makefile",
+      "LOC=-DLARGE_FILE_SUPPORT -DSYMLINKS",
+      "D_USE_BZ2=-DUSE_BZIP2",
+      "IZ_BZIP2=\"#{Formula["bzip2"].opt_prefix}/lib\"",
```
DomT4: `opt_lib` will work fine. You can use single quotes here as well inside the `""` to avoid the slightly gross escaping in `make` arguments.

unzip.rb
felixbuenemann: I did get compile error with the system linbz2, although it's the same version as the one in dupes. I could re-evaluate what was going wrong.

unzip.rb
felixbuenemann: Sorry, totally forgot about bumping the revision. I didn't know about the patches repo, I'll look into it.

unzip.rb
DomT4: If we can avoid the extra mandatory dependency that's desirable.

felixbuenemann: Btw. the libbz2 in dupes does not install shared libs (not sure why), so technically this is a build only dependency…

DomT4: May well be off by default and nobody has ever asked to change that, I suspect. My objection is less if it's a pure `:build` dep only.

unzip.rb
felixbuenemann: Will switch to external patch, once Homebrew/formula-patches#23 is merged.

DomT4: Merged there, Thanks!

felixbuenemann:
@DomT4 I figured out the compile problem it was related to some missing headers.

The formula looks much cleaner now, thanks for the feedback.

felixbuenemann:
@DomT4 Somehow the jenkins build seems stuck, the yosemite job is referencing build 280 instead of build 281.

DomT4:
Seems okay?

felixbuenemann:
Yeah, seems fine now, not sure why it stalled so long on the yosemite build.

unzip.rb
```diff
@@ -4,11 +4,25 @@ class Unzip < Formula
   url "https://downloads.sourceforge.net/project/infozip/UnZip%206.x%20%28latest%29/UnZip%206.0/unzip60.tar.gz"
   version "6.0"
   sha256 "036d96991646d0449ed0aa952e4fbe21b476ce994abc276e49d30e686708bd37"
+  revision 1
 
   keg_only :provided_by_osx
 
+  # Upstream is unmaintained so we use the Ubuntu unzip-6.0-20ubuntu1 patchset:
```
MikeMcQuaid: If upstream is unmaintained we should boneyard this formula.

felixbuenemann: It is maintained by Debian for years. Unfortunately the version that ships with Mac OS X is very outdated and lacks important features like ZIP64 support, so the dupe is useful. There programs and scripts that depend on unzip and also on features not provided by the older OS X version, so dropping it would be a bad idea.

DomT4: I'm reasonably okay with this, despite the enormity of the patch. Dupes does exist to provide "better" (read into "better" whatever you wish to) versions of system utilities, and as long as people are willing to try and keep on top of patch situation I don't mind that here.

ilovezfs: I'd be more comfortable with this if we were using pseudo-upstream URLs for the patching. For example, see https://github.com/Homebrew/homebrew-core/blob/master/Formula/p7zip.rb#L15-L24

(Also, note http://pkgs.fedoraproject.org/cgit/rpms/unzip.git/tree/)

And why Ubuntu not Debian? :trollface:

Why hasn't someone forked this? Is there no GitHub project with the patches integrated and tagged releases?

MikeMcQuaid: > Why hasn't someone forked this? Is there no GitHub project with the patches integrated and tagged releases?

This is what concerns me too.

DomT4: I suspect that the Linux distros (which effectively _have_ forked `unzip` by this point) doing so more or less kills a lot of the need/desire for someone independent to do the work themselves.

ilovezfs: Debian:
```
01-manpages-in-section-1-not-in-section-1l
02-branding-patch-this-is-debian-unzip
03-include-unistd-for-kfreebsd
04-handle-pkware-verification-bit
05-fix-uid-gid-handling
06-initialize-the-symlink-flag
07-increase-size-of-cfactorstr
08-allow-greater-hostver-values
09-cve-2014-8139-crc-overflow
10-cve-2014-8140-test-compr-eb
11-cve-2014-8141-getzip64data
12-cve-2014-9636-test-compr-eb
13-remove-build-date
14-cve-2015-7696
15-cve-2015-7697
16-fix-integer-underflow-csiz-decrypted
```
To which Ubuntu adds:
```

```

ilovezfs: Fedora:
```
unzip-6.0-bzip2-configure.patch
unzip-6.0-close.patch
unzip-6.0-exec-shield.patch
unzip-6.0-attribs-overflow.patch
unzip-6.0-nostrip.patch
unzip-6.0-manpage-fix.patch
unzip-6.0-fix-recmatch.patch
unzip-6.0-symlink.patch
unzip-6.0-caseinsensitive.patch
unzip-6.0-format-secure.patch
unzip-6.0-valgrind.patch
unzip-6.0-x-option.patch
unzip-6.0-overflow.patch
unzip-6.0-cve-2014-8139.patch
unzip-6.0-cve-2014-8140.patch
unzip-6.0-cve-2014-8141.patch
unzip-6.0-overflow-long-fsize.patch
unzip-6.0-heap-overflow-infloop.patch
unzip-6.0-alt-iconv-utf8.patch
unzip-6.0-alt-iconv-utf8-print.patch
```
Gentoo (which also has the Debian patchset)
```
 	unzip-5.52-CVE-2008-0888.patch
	unzip-5.52-no-exec-stack.patch
	unzip-6.0-natspec.patch
	unzip-6.0-no-exec-stack.patch
```

Arch:
```
overflow-fsize.patch
cve20149636.patch
test_compr_eb.patch
getZip64Data.patch
crc32.patch
empty-input.patch
csiz-underflow.patch
nextbyte-overflow.patch
```
pkgsrc:
```
patch-ab
patch-ac
patch-crypt.c
patch-extract.c
patch-fileio.c
patch-list.c
patch-process.c
patch-unix_unxcfg.h    
```

ilovezfs: I'm not sure I'd call that a coherent fork, but certainly everyone is at least patching the CVEs.

felixbuenemann: Personally I'm unopinionated about which patchset to choose. I used the ubuntu one because it had an additional patch on top of the upstream debian package. The fedora patches could be linked to directly, so no separate patchset would be needed (I used `for f in $(cat series); do cat $f; done > patchset` with the debian archives).

ilovezfs: @felixbuenemann I'd suggest this:
```
class Unzip < Formula
  desc "Extraction utility for .zip compressed archives"
  homepage "http://www.info-zip.org/pub/infozip/UnZip.html"
  url "https://downloads.sourceforge.net/project/infozip/UnZip%206.x%20%28latest%29/UnZip%206.0/unzip60.tar.gz"
  version "6.0"
  sha256 "036d96991646d0449ed0aa952e4fbe21b476ce994abc276e49d30e686708bd37"

  keg_only :provided_by_osx

  patch do
    url "https://mirrors.ocf.berkeley.edu/debian/pool/main/u/unzip/unzip_6.0-20.debian.tar.xz"
    mirror "https://mirrors.kernel.org/debian/pool/main/u/unzip/unzip_6.0-20.debian.tar.xz"
    sha256 "7ddfafb1a771ae8d6b4e25c5e31f22c717e0fe606b1bafadcdd574c01f671490"
    apply %w[
             patches/01-manpages-in-section-1-not-in-section-1l
             patches/02-branding-patch-this-is-debian-unzip
             patches/03-include-unistd-for-kfreebsd
             patches/04-handle-pkware-verification-bit
             patches/05-fix-uid-gid-handling
             patches/06-initialize-the-symlink-flag
             patches/07-increase-size-of-cfactorstr
             patches/08-allow-greater-hostver-values
             patches/09-cve-2014-8139-crc-overflow
             patches/10-cve-2014-8140-test-compr-eb
             patches/11-cve-2014-8141-getzip64data
             patches/12-cve-2014-9636-test-compr-eb
             patches/13-remove-build-date
             patches/14-cve-2015-7696
             patches/15-cve-2015-7697
             patches/16-fix-integer-underflow-csiz-decrypted
          ]
  end

  def install
    system "make", "-f", "unix/Makefile",
      "CC=#{ENV.cc}",
      "LOC=-DLARGE_FILE_SUPPORT -DSYMLINKS",
      "D_USE_BZ2=-DUSE_BZIP2",
      "L_BZ2=-lbz2",
      "LFLAGS1=-liconv",
      "macosx"
    system "make", "prefix=#{prefix}", "MANDIR=#{man}", "install"
  end

  test do
    system "#{bin}/unzip", "--help"
  end
end

```

felixbuenemann: That's nice, I didn't know you could apply multiple patches from an archive!

ilovezfs: :) Yeah, it was motivated by the rsync patches, which rsync upstream itself distributes as a single tarball of goodies of which we use three.

felixbuenemann:
I pushed a new version that uses the patch archive and `apply` solution suggested by @ilovezfs above. I also dropped the `SYMLINKS` define, because it was enabled anyways.

unzip.rb
```diff
@@ -4,11 +4,45 @@ class Unzip < Formula
...
+  # http://changelogs.ubuntu.com/changelogs/pool/main/u/unzip/unzip_6.0-20ubuntu1/changelog
+  patch do
+    url "https://launchpad.net/ubuntu/+archive/primary/+files/unzip_6.0-20ubuntu1.debian.tar.xz"
+    mirror "http://archive.ubuntu.com/ubuntu/pool/main/u/unzip/unzip_6.0-20ubuntu1.debian.tar.xz"
```
ilovezfs: if there's an https mirror you can find, that would be even better

ilovezfs: @felixbuenemann here's one https://www.mirrorservice.org/sites/archive.ubuntu.com/ubuntu/pool/main/u/unzip/unzip_6.0-20ubuntu1.debian.tar.xz

DomT4:
Merged in https://github.com/Homebrew/homebrew-dupes/commit/faeb12a2f67d9edd82bdebad11f8383c5f217594. Thank you for your contribution to Homebrew @felixbuenemann; we appreciate it! :smiley_cat:

ilovezfs:
Yeah thank you for this @felixbuenemann. Despite the unorthodox upstream situation, I'm happy with how it turned out since it patches CVEs that affect Apple's version, too.

PR #615, opened 2016-05-29
ncurses: Symlink libtinfo.so and ncurses.pc

sjackman:
Symlink `libncurses.so` to `libtinfo.so`
Symlink `ncursesw.pc` to `ncurses.pc`


sjackman:
See https://github.com/Homebrew/homebrew-dupes/pull/517 for `libtinfo.so`
See https://github.com/Linuxbrew/homebrew-core/pull/254#issuecomment-218944939 for `ncurses.pc`

ncurses.rb
```diff
@@ -69,9 +69,14 @@ def make_libncurses_symlinks
         ln_s "libncurses.dylib", "libcurses.dylib"
       else
         ln_s "libncurses.so", "libcurses.so"
+        ln_s "libncurses.so", "libtinfo.so"
       end
     end
 
+    cd lib/"pkgconfig" do
+      ln_s "ncursesw.pc", "ncurses.pc"
```
DomT4: Should this be Linux exclusive or? I vaguely recall some discussion about this in the past amongst Homebrew maintainers but I can't remember what was actually said now.

DomT4: You may also be able to condense this a bit by doing something like:

```ruby
(lib/"pkgconfig").install_symlink lib/"pkgconfig/ncursesw.pc" => "ncurses.pc"
```

sjackman: The rest of this formula uses `ln_s` rather than `install_symlink`. Consistency within the formula looks better, I think. They could all be changed to `install_symlink`, if that's preferable.

sjackman: It's relatively uncommon to require `ncurses.pc`. I've only run into one formula so far that needs it: `infer`.

DomT4: > The rest of this formula uses ln_s rather than install_symlink. Consistency within the formula looks better, I think. They could all be changed to install_symlink, if that's preferable.

We really need to harmonise usage throughout Homebrew but `.install_symlink` is preferable generally.

ilovezfs: what is the difference between them?

DomT4: https://github.com/Homebrew/brew/blob/341b64cc962f8ff69a37d7e20f6d94c7d5658ca7/share/doc/homebrew/Formula-Cookbook.md#file-level-operations

sjackman: >>The rest of this formula uses ln_s rather than install_symlink. Consistency within the formula looks better, I think. They could all be changed to install_symlink, if that's preferable.

>We really need to harmonise usage throughout Homebrew but .install_symlink is preferable generally.

So for this PR, leave it as is, or change all of them?

DomT4: > or change all of them?

Ideally, this one 😉.

sjackman: Done!

DomT4:
Seems alright. Merged in https://github.com/Homebrew/homebrew-dupes/commit/0c94c6ffed660cd94cd8758405c0968d76cf1800 & https://github.com/Homebrew/homebrew-dupes/commit/5b59549bfa0678ffc1f73145cf5cc639f79a21ea. Thanks again @sjackman! 😃

sjackman:
Thanks, Dominyk!

PR #601, opened 2016-05-05
lsof 4.89

sjackman:
Both previous mirrors no longer work.

PR #537, opened 2015-12-18
lsof 4.89

crlf0710:

PR #616, opened 2016-05-30
lsof 4.89 with compile fix

felixbuenemann:
See comments from @ilovezfs on #601 and also #537.

I had to get a little creative with the patch, because the lsof source is stored as a tar inside the source archive, so the normal patch process can't be used. Instead the patch first creates another patch file, which is then applied after unpacking the source tar. I think that's still cleaner than using `inreplace`.

### DIFF
--- a/lsof.rb
+++ b/lsof.rb
@@ -1,11 +1,17 @@
+class LsofDownloadStrategy < CurlDownloadStrategy
+  def stage
+    super
+    safe_system "/usr/bin/tar", "xf", "#{name}_#{version}_src.tar"
+    cd "#{name}_#{version}_src"
+  end
+end
+
 class Lsof < Formula
   desc "Utility to list open files"
   homepage "https://people.freebsd.org/~abe/"
-  # 4.89 has major problems on OS X blocking upgrade for now.
-  # https://github.com/Homebrew/homebrew-dupes/pull/537
-  # url "ftp://sunsite.ualberta.ca/pub/Mirror/lsof/lsof_4.88.tar.bz2"
-  url "https://mirrorservice.org/sites/lsof.itap.purdue.edu/pub/tools/unix/lsof/OLD/lsof_4.88.tar.bz2"
-  sha256 "fe6f9b0e26b779ccd0ea5a0b6327c2b5c38d207a6db16f61ac01bd6c44e5c99b"
+  url "https://mirrorservice.org/sites/lsof.itap.purdue.edu/pub/tools/unix/lsof/lsof_4.89.tar.bz2",
+    :using => LsofDownloadStrategy
+  sha256 "81ac2fc5fdc944793baf41a14002b6deb5a29096b387744e28f8c30a360a3718"
 
   bottle do
     cellar :any_skip_relocation
@@ -14,22 +20,26 @@ class Lsof < Formula
     sha256 "92a1f687aa8a0df343b90771e16f6a11aa60968c034c9660318e38142533fb82" => :mavericks
   end
 
+  patch do
+    url "https://raw.githubusercontent.com/Homebrew/formula-patches/c3acbb8/lsof/lsof-489-darwin-compile-fix.patch"
+    sha256 "997d8c147070987350fc12078ce83cd6e9e159f757944879d7e4da374c030755"
+  end
+
   def install
     ENV["LSOF_INCLUDE"] = "#{MacOS.sdk_path}/usr/include"
 
-    system "tar", "xf", "lsof_#{version}_src.tar"
+    # Source hardcodes full header paths at /usr/include
+    inreplace %w[
+      dialects/darwin/kmem/dlsof.h
+      dialects/darwin/kmem/machine.h
+      dialects/darwin/libproc/machine.h
+    ], "/usr/include", "#{MacOS.sdk_path}/usr/include"
 
-    cd "lsof_#{version}_src" do
-      # Source hardcodes full header paths at /usr/include
-      inreplace %w[dialects/darwin/kmem/dlsof.h dialects/darwin/kmem/machine.h
-                   dialects/darwin/libproc/machine.h],
-                "/usr/include", MacOS.sdk_path.to_s + "/usr/include"
-
-      mv "00README", "../README"
-      system "./Configure", "-n", `uname -s`.chomp.downcase
-      system "make"
-      bin.install "lsof"
-    end
+    mv "00README", "README"
+    system "./Configure", "-n", `uname -s`.chomp.downcase
+    system "make"
+    bin.install "lsof"
+    man8.install "lsof.8"
   end
 
   test do


lsof.rb
```diff
@@ -1,11 +1,10 @@
...
   homepage "https://people.freebsd.org/~abe/"
-  # 4.89 has major problems on OS X blocking upgrade for now.
-  # https://github.com/Homebrew/homebrew-dupes/pull/537
-  # url "ftp://sunsite.ualberta.ca/pub/Mirror/lsof/lsof_4.88.tar.bz2"
-  url "https://mirrorservice.org/sites/lsof.itap.purdue.edu/pub/tools/unix/lsof/OLD/lsof_4.88.tar.bz2"
-  sha256 "fe6f9b0e26b779ccd0ea5a0b6327c2b5c38d207a6db16f61ac01bd6c44e5c99b"
+  url "https://mirrorservice.org/sites/lsof.itap.purdue.edu/pub/tools/unix/lsof/lsof_4.89.tar.bz2"
+  sha256 "81ac2fc5fdc944793baf41a14002b6deb5a29096b387744e28f8c30a360a3718"
+
+  patch :DATA
```
DomT4: It possibly makes sense to submit this as a patch to the patches repo again, and then use that as a `resource` and then stage the `resource` into the `buildpath` and apply it. Also needs some explanation as to what it is fixing, where the upstream report is, etc etc.

felixbuenemann:
@DomT4 I sent a PR to the patches repo.

DomT4:
@felixbuenemann Merged there, Thanks! Any chance that mail was to a public mailing list at all?

felixbuenemann:
No, it was the mail mentioned on the homepage.

felixbuenemann:
@DomT4 OK, updated to use the patch as a resource.

felixbuenemann:
@DomT4 Fixes should be included in the next lsof version whenever that'll be:

> Felix,
> 
> I have installed your patches in the lsof source, but I have no
> schedule for releasing the next lsof revision.
> 
> Nor do I test on Mac OS X any more.
> 
> Regards,
> 
> Vic

DomT4:
@ilovezfs Are you happy here? You proposed the original idea so happy to let you yay/nay it.

ilovezfs:
I'm confused. Why aren't we using patch-do?

felixbuenemann:
Because we would have to patch a tar archive in that case…

ilovezfs:
I think we can define a custom download strategy in the formula to work around that.

felixbuenemann:
If you can provide an example I'm happy to change that.

ilovezfs:
writing one for you now ;)

DomT4:
> I think we can define a custom download strategy in the formula to work around that.

Blurgh, custom download strategies are gross 😜. As long as we stick some giant notation above it in terms of "Please remove on next update" it's reasonably okay.

ilovezfs:
it's actually dramatically simplified things ;)

ilovezfs:
@DomT4 @felixbuenemann here it is
```
class LsofDownloadStrategy < CurlDownloadStrategy
  def stage
    safe_system "/usr/bin/tar", "-xjf", cached_location
    nv = "#{name}_#{version}"
    safe_system "/usr/bin/tar", "xf", "#{nv}/#{nv}_src.tar"
    rm_r nv
    cd "#{nv}_src"
  end
end

class Lsof < Formula
  desc "Utility to list open files"
  homepage "https://people.freebsd.org/~abe/"
  url "https://mirrorservice.org/sites/lsof.itap.purdue.edu/pub/tools/unix/lsof/lsof_4.89.tar.bz2",
      :using => LsofDownloadStrategy
  sha256 "81ac2fc5fdc944793baf41a14002b6deb5a29096b387744e28f8c30a360a3718"

  bottle do
    cellar :any_skip_relocation
    sha256 "166741406a0a1d6ac78b82274d56f1bb1d24c4d4cd2421919e48ce42566f90ca" => :el_capitan
    sha256 "b8292349936eea9e0f24daf01818cba06ce5e689b4e97f026a431034c509e67d" => :yosemite
    sha256 "92a1f687aa8a0df343b90771e16f6a11aa60968c034c9660318e38142533fb82" => :mavericks
  end

  patch do
    url "https://raw.githubusercontent.com/Homebrew/formula-patches/c3acbb8/lsof/lsof-489-darwin-compile-fix.patch"
    sha256 "997d8c147070987350fc12078ce83cd6e9e159f757944879d7e4da374c030755"
  end

  def install
    ENV["LSOF_INCLUDE"] = "#{MacOS.sdk_path}/usr/include"
    # Source hardcodes full header paths at /usr/include
    inreplace %w[dialects/darwin/kmem/dlsof.h dialects/darwin/kmem/machine.h
                 dialects/darwin/libproc/machine.h],
              "/usr/include", MacOS.sdk_path.to_s + "/usr/include"
    
      mv "00README", "../README"
      system "./Configure", "-n", `uname -s`.chomp.downcase
      system "make"
      bin.install "lsof"
  end

  test do
    (testpath/"test").open("w") do
      system "#{bin}/lsof", testpath/"test"
    end
  end
end
```

ilovezfs:
@DomT4 Also, I wouldn't get rid of it at next release, since presumably upstream will continue to have their funky packaging, and this makes it "normal."

DomT4:
I'd rather we didn't carry custom download strategies around forever, to be honest. A fair number of contributors find them hard to understand, and it is essentially a hack.

ilovezfs:
What exactly do you object to in what's above other than "it's a hack"? It

a) avoids unpacking tarballs in the the install method
b) should continue to work for this formula in the future
c) reduces indentation of the install method
d) allows patch-do to work
e) hard codes nothing
f) improves readability of the install method
g) has precedent in core formulae

DomT4:
Custom download mechanisms, by the fact they are _custom download mechanisms_ and our normal packaging process doesn't work for them, implies there's a packaging issue upstream we're having to work around and is something upstream should be nudged to resolve.

> has precedent in core formulae

There are what, 3600 formulae in the core? And no more than a handful have custom download strategies. We have, to date, more or less treated custom download strategies like a bug that needs to be squashed.

ilovezfs:
Turns out I was able to make it even simpler:
```
class LsofDownloadStrategy < CurlDownloadStrategy
  def stage
    super
    safe_system "/usr/bin/tar", "xf", "#{name}_#{version}_src.tar"
    cd "#{name}_#{version}_src"
  end
end
```

felixbuenemann:
I lean towards using the download strategy here. At the end of the day both the dl strategy and resouce vs. patch work around non-standard packaging, so you are trading one hack for another. I wouldn't get my hopes up on getting that fixed upstream.

ilovezfs:
Seriously, `super` plus two lines and `patch do`. You should be smiling!

>We have, to date, more or less treated custom download strategies like a bug that needs to be squashed.

I think that's exactly the situation here, actually. Upstream should distribute it as a tarball link + signature file link like every other sane packaging does, not wrap a tarball in a tarball with a dangling signature file and a readme inside the tarball and another readme inside the tarball inside the tarball. This upstream *should* be bugged to make this unnecessary.

DomT4:
> You should be smiling!

What is this smiling thing you talk about?

> This upstream should be bugged to make this unnecessary.

Yup. I'll take this temporarily but my position, which I admit may be harder than other maintainers, is that like any bug that won't get fixed by upstream and makes packaging more intensive for us merits a discussion about 💀 eventually.

ilovezfs:
>What is this smiling thing you talk about?

I believe it is a physical hint of the affect of certain emoticons. Or at least that's what I'm told.

>Yup. I'll take this temporarily 

:cake: :icecream: :balloon:

>like any bug that won't get fixed by upstream

It doesn't hurt to ask.

ilovezfs:
Regarding the underlying patch, I'm happy with it since upstream didn't balk at the changes. If there are problems, we can switch to the punting method, `PROC_FP_GUARDED  -> DOGS_ARE_THE_BEST`.

ilovezfs:
@DomT4 BTW, feel free to veto the custom download strategy if you'd like. I won't be upset. I just considered this my penance for 
https://github.com/Homebrew/homebrew-core/commit/3bea9d40701a763fe3e389ecca42fc94315ab3ff#diff-0102adc69deed238616ca71311d8ee85R50
and
https://github.com/Homebrew/homebrew-core/commit/022674375c5ad0bffca0e5211fb16335e6381245#diff-a774800229c8c00a934f28bdcdf51676R55

DomT4:
Na, it's fine temporarily, my "beef" is only with carrying them around long-term when there's packaging fixes upstream could and should make to negate the need for a CDS.

ilovezfs:
BTW another option might be to always unpack top level tarballs.

ilovezfs:
Looks like that wouldn't benefit any of the cases of `system "tar", ...` in core, so I guess not worth it.

felixbuenemann:
Updated to use the custom download strategy.

lsof.rb
```diff
@@ -14,22 +20,23 @@ class Lsof < Formula
...
+    inreplace %w[dialects/darwin/kmem/dlsof.h dialects/darwin/kmem/machine.h
+                 dialects/darwin/libproc/machine.h],
+              "/usr/include", MacOS.sdk_path.to_s + "/usr/include"
 
-      mv "00README", "../README"
-      system "./Configure", "-n", `uname -s`.chomp.downcase
-      system "make"
-      bin.install "lsof"
-    end
+    mv "00README", "../README"
```
felixbuenemann: This looks weird, shouldn't it move the README outside the build dir? I'll check that.

ilovezfs: This has to be changed to `mv "00README", "README"`.

ilovezfs: Yeah I had to change that if you check the version I pasted: https://github.com/Homebrew/homebrew-dupes/pull/616#issuecomment-223257342

felixbuenemann: Yeah, and we should install the manpage, I'll fix that.

ilovezfs: Cool, thanks.

felixbuenemann:
@DomT4 Should be good to merge now.

lsof.rb
```diff
@@ -14,22 +20,24 @@ class Lsof < Formula
...
-
-    cd "lsof_#{version}_src" do
-      # Source hardcodes full header paths at /usr/include
-      inreplace %w[dialects/darwin/kmem/dlsof.h dialects/darwin/kmem/machine.h
-                   dialects/darwin/libproc/machine.h],
-                "/usr/include", MacOS.sdk_path.to_s + "/usr/include"
+    # Source hardcodes full header paths at /usr/include
+    inreplace %w[dialects/darwin/kmem/dlsof.h dialects/darwin/kmem/machine.h
+                 dialects/darwin/libproc/machine.h],
+              "/usr/include", MacOS.sdk_path.to_s + "/usr/include"
```
ilovezfs: Is that idiomatic as-is spanning multiple lines? Looks a little odd to me. Maybe:
```
    %w[
      dialects/darwin/kmem/dlsof.h
      dialects/darwin/kmem/machine.h
      dialects/darwin/libproc/machine.h
    ].each do |f| 
      inreplace f, "/usr/include", MacOS.sdk_path.to_s + "/usr/include"
    end
```
@DomT4 thoughts?

felixbuenemann:
```ruby
    inreplace %w[
      dialects/darwin/kmem/dlsof.h
      dialects/darwin/kmem/machine.h
      dialects/darwin/libproc/machine.h
    ], "/usr/include", MacOS.sdk_path.to_s + "/usr/include"
```
Would be shorter and just as readable.

ilovezfs: yeah I just don't know what @DomT4's stylistic preference is here. Usually when %w[ spans multiple lines you align the words with the opening bracket [ and align the closing bracket ] with the %.

felixbuenemann: Well, I pushed my solution. We're probably in esoteric territory right now… 😄

ilovezfs: You'd be surprised ;)

felixbuenemann: Btw. we should probably move those fixes into the patch.

ilovezfs: You won't know `MacOS.sdk_path.to_s` until runtime.

felixbuenemann: And we don't need to know, see 3d6845a.

ilovezfs: Why is it even needed at all?

felixbuenemann: Because the source hardcodes include paths, which it shouldn't.

ilovezfs: Why do we care? It compiles fine without the replacements.

felixbuenemann: But probably only, if CLI tools are installed.

ilovezfs: Nope, it's fine without them.

felixbuenemann: In that case we should also drop the `LSOF_INCLUDE` environment variable, because the point of the `inreplace` was to make is use the includes from the given SDK.

ilovezfs: The problem was https://github.com/Homebrew/homebrew-dupes/issues/403

I just tested on 10.10 without the CLT, without the replaces, and with and without the LSOF_INCLUDE. No problems compiling. So it looks like

https://github.com/Homebrew/homebrew-dupes/issues/398
https://github.com/Homebrew/homebrew-dupes/commit/9b8998d3664877dd8703b4b2f376b4a0cb6cc1c0

https://github.com/Homebrew/homebrew-dupes/issues/403
https://github.com/Homebrew/homebrew-dupes/commit/612d10556c16eda30ab0306cf6c18c360c3080d8

were both fixed upstream at some point.

felixbuenemann: How do you uninstall xcode command line tools? I'd like to try locally.

ilovezfs: It's possible a superenv changed fixed it, but it's also possible it's still broken on 10.9 without the CLT. So until someone tests that specifically, I'd just leave it alone.

DomT4:
I only went out for 2 hours to do some shopping & you've left me 20+ emails to return to 😄. I'll look through & see what I need to reply to.

felixbuenemann:
Lol, just read the last commit 😄

lsof.rb
ilovezfs: To uninstall CLT on 10.10 and 10.11 you can drag or mv /Library/Developer/CommandLineTools to the Trash (~/.Trash) temporarily. For <= 10.9, you actually have to go remove the files listed by pkgutil --lsbom for the CLT package id.

ilovezfs:
>I only went out for 2 hours to do some shopping

Well now you know not to do that. Shame, shame

lsof.rb
DomT4: > Is that idiomatic as-is spanning multiple lines? Looks a little odd to me. Maybe:

ILZ's suggestion here on style makes me happy on readability, but it's not a massive deal.

DomT4:
> Well now you know not to do that.

Sadly Homebrew alone doesn't quench my desire to sit around drinking whiskey sours in the sun 😉.

This looks basically fine to me. Nice work both 👍.

lsof.rb
```diff
@@ -14,22 +20,26 @@ class Lsof < Formula
...
-      # Source hardcodes full header paths at /usr/include
-      inreplace %w[dialects/darwin/kmem/dlsof.h dialects/darwin/kmem/machine.h
-                   dialects/darwin/libproc/machine.h],
-                "/usr/include", MacOS.sdk_path.to_s + "/usr/include"
+    # Source hardcodes full header paths at /usr/include
+    inreplace %w[
+      dialects/darwin/kmem/dlsof.h
+      dialects/darwin/kmem/machine.h
+      dialects/darwin/libproc/machine.h
+    ], %r{"/usr/include/([^"]+)"}, '<\1>'
```
ilovezfs: This should be reverted to the way it was.

ilovezfs:
@felixbuenemann Yeah, let's go with
```
    %w[
      dialects/darwin/kmem/dlsof.h
      dialects/darwin/kmem/machine.h
      dialects/darwin/libproc/machine.h
    ].each do |f| 
      inreplace f, "/usr/include", MacOS.sdk_path.to_s + "/usr/include"
    end
```
and then I think this can be pulled.

lsof.rb
felixbuenemann: Why?

ilovezfs: Because it precisely matches `"#{MacOS.sdk_path}/usr/include"` set for LSOF_INCLUDE above, and because if it ain't broke, don't fix it.

felixbuenemann: OK, I reverted the regex part. I still think the internal iterator solution is nicer than using the external iterator, so I left that part alone.

felixbuenemann:
@ilovezfs Are you OK with 141f385 or should I also switch to the `each` version?

DomT4:
@ilovezfs Feel free to :shipit: when you're 👍  here.

ilovezfs:
@felixbuenemann Thanks for your work on this and sorry for the delay! :bow:

PR #617, opened 2016-06-02
[nano] updating HEAD to git repo

tmaone:
Nano's source tree has been migrated to [`git`](http://www.nano-editor.org/git.php).

### DIFF
--- a/nano.rb
+++ b/nano.rb
@@ -13,7 +13,7 @@ class Nano < Formula
   end
 
   head do
-    url "git://git.savannah.gnu.org/nano.git"
+    url "http://git.savannah.gnu.org/r/nano.git"
     depends_on "automake" => :build
     depends_on "autoconf" => :build
   end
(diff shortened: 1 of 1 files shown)


nano.rb
```diff
@@ -13,7 +13,7 @@ class Nano < Formula
   end
 
   head do
-    url "svn://svn.sv.gnu.org/nano/trunk/nano"
+    url "git://git.savannah.gnu.org/nano.git"
```
DomT4: Any chance `http://git.savannah.gnu.org/r/nano.git` works instead here? The `git://` protocol has a bit of a habit of getting taken out by restrictive firewalls.

tmaone: Just checked. It works fine. I'll update to `http` then.

DomT4:
Merged in https://github.com/Homebrew/homebrew-dupes/commit/7e4f13df8db5c01d04b96abb95c91aec884d047d. Thank you for your contribution to Homebrew @tmaone; we appreciate it! :smiley_cat:

PR #620, opened 2016-06-13
unzip: Depends on bzip2 and fixes for Linuxbrew

sjackman:


unzip.rb
```diff
@@ -43,13 +45,15 @@ class Unzip < Formula
...
+    args = []
+    args << "LFLAGS1=-liconv" if OS.mac?
     system "make", "-f", "unix/Makefile",
       "CC=#{ENV.cc}",
       "LOC=-DLARGE_FILE_SUPPORT",
       "D_USE_BZ2=-DUSE_BZIP2",
       "L_BZ2=-lbz2",
-      "LFLAGS1=-liconv",
-      "macosx"
+      "macosx",
```
DomT4: Is `macosx` not problematic for Linux, or?

sjackman: Believe it or not, the `macosx` build rule is actually much saner than the `linux` build rule, which is weird. In fact, that whole Makefile is one of the more odd Makefiles that I've ever seen in my life. Short answer though, the `macosx` target works just fine on Linux.

DomT4: Heh, oh well 🙈. Thanks for the answer!

DomT4:
Merged in https://github.com/Homebrew/homebrew-dupes/commit/9e4a084bdcfa435c5a00e6b5df4961c1809c988a. Thanks @sjackman!

sjackman:
Thanks for the merge, Dominyk!

PR #621, opened 2016-06-13
openssh: Depends on libedit and krb5 for Linuxbrew

sjackman:
Remove pam dependency for Linuxbrew.
Fix mkdir: cannot create directory `/var/empty': Permission denied


DomT4:
Merged in https://github.com/Homebrew/homebrew-dupes/commit/a892bf44a80d6358be5133662ecb3adc90a7338b. Thanks again Shaun.

sjackman:
Thanks once again for the merge, Dominyk!

PR #624, opened 2016-06-23
ed: Symlink ed to ged with --default-names

sjackman:
Symlink `ed` to `ged` when compiled with `--default-names` to fix `test`.


ilovezfs:
This doesn't seem like the right way to fix the bug. If I'm using default names, there shouldn't be any g+ prefix around at all. Is this the bug where superenv fails to pick the brew version when it has the default name? Or is it a different bug?

sjackman:
`brew test` fails because it executes `ged`, which is named `ed` if you install it `--with-default-names`.

Any scripts that depend on GNU `ed` being named `ged` on Homebrew will fail if it's installed `--with-default-names`. This fix ensures that those scripts will continue to work.

sjackman:
```
Error: 1 problem in 1 formula
homebrew/dupes/ed:
  * Stable: Please use https:// for http://ftpmirror.gnu.org/ed/ed-1.12.tar.lz
```
http://bot.brew.sh/job/Homebrew%20Dupes%20Pull%20Requests/311/version=el_capitan/testReport/junit/brew-test-bot/el_capitan/audit_homebrew_dupes_ed/

ilovezfs:
@sjackman I think I'd change the test to check the build options and then execute the correct binary. For example, https://github.com/Homebrew/homebrew-core/blob/master/Formula/idris.rb#L56-L63

sjackman:
Done. Also fixed the above audit warning.

sjackman:
@ilovezfs Bump. =)

ilovezfs:
@sjackman looks good. You might also want to add a third parameter to the `pipe_output`s asserting the exit status is zero, since `pipe_output` won't do that by default, unlike `shell_output`, which does, though it's hard to imagine the replacements succeeding without the exit status having been zero, but you never know!

sjackman:
Fixed. Thanks for the suggestion, Joe.

ilovezfs:
:+1:

ed.rb
```diff
@@ -37,7 +37,11 @@ def caveats
   test do
     testfile = (testpath/"test")
```
ilovezfs: This doesn't actually need the parentheses if you feel like removing them

sjackman: Fixed.

ilovezfs:
@DomT4 :ship:?

ilovezfs:
@sjackman I guess to be consistent, this seems to be the idiom:
https://github.com/Homebrew/homebrew-dupes/blob/1542997258b0fca605c95a3df23ea200cbb9bff5/grep.rb#L50
https://github.com/Homebrew/homebrew-dupes/blob/046ed555453d7f464c044d2da0f83764c614feaf/make.rb#L37

DomT4:
Prefer the way ILZ points out above, but in theory 👍.

sjackman:
Agreed. I recall that I have had pushback against using the ternary operator in formulas, but I like it just fine. Fixed.

DomT4:
Mike's not a fan of them, but I think handling default names in tests is a reasonable use case.

sjackman:
If it's for readability, (because `?` and `:` look weird), I quite like using if/then/else/end, though it's more verbose.
```ruby
cmd = build.with?("default-names") ? "ed" : "ged"
cmd = if build.with?("default-names") then "ed" else "ged" end
```

DomT4:
It's more that ternary operators can be obtuse to people new to Ruby, and formulae are _supposed_ to be fairly easy.

ilovezfs:
I think we can expect "them" to learn the ternary operator. If we need to use stick figures and crayons to help, I'm not opposed.

ilovezfs:
@BrewTestBot test this please

ilovezfs:
pulled.

sjackman:
Thanks, Joe.

Issue #622, opened 2016-06-13
ncurses fails to build on RedHatEnterpriseServer 6.7 


nickjhathaway:
Following instructions here,  
https://github.com/Linuxbrew/linuxbrew/wiki/Standalone-Installation,  
I'm trying to install a standalone installation of linuxbrew on a RedHatEnterpriseServer 6.7 cluster. I'm running into an issue with homebrew/dupes/ncurses and the gist of the issue is here,  
https://gist.github.com/anonymous/3d5a6fa1fcb71d7d55f0afbabcc7b6b3#file-02-make-L2-L1440  
I wish i could be more specific but I'm having trouble figuring out what the error even is. What I'm running into is, 

```
...
...
/home/nh79w/.linuxbrew/bin/ginstall -c -m 644 ../lib/libncursesw_g.a /home/nh79w/.linuxbrew/Cellar/ncurses/6.0_1/lib/libncursesw_g.a
/home/nh79w/.linuxbrew/bin/ginstall: cannot stat '../lib/libncursesw_g.a': No such file or directory
...
...
```

And I'm not sure why that file isn't there for ginstall but it's the first error message i can find in the logs

I've ran brew update twice and haven't been able to find a similar issue already open.

DomT4:
CC @sjackman.

sjackman:
If you can install in `/home/linuxbrew/.linuxbrew` you can make use of precompiled binary bottles as a workaround.

Install errors like that are sometimes a race condition. Try:
```sh
HOMEBREW_MAKE_JOBS=1 brew install homebrew/dupes/ncurses
```

nickjhathaway:
Excellent! 
```
HOMEBREW_MAKE_JOBS=1 brew install homebrew/dupes/ncurses
```
Worked and ncurses was able to install properly, thanks for the help!

sjackman:
@DomT4 This should fix it:
```ruby
system "make"
ENV.deparallelize
system "make", "install"
```
With 32–256 threads I've noticed a number of race conditions that don't show up on desktop machines.

DomT4:
@sjackman Mind filing a PR? Possibly worth nudging upstream as well. Things _should_ build in parallel, in an ideal world.

sjackman:
Opened a PR at #630.

PR #630, opened 2016-07-02
ncurses: Deparallelize make install

sjackman:
Use https for brew audit.

Fixes Homebrew/homebrew-dupes#622.


dunn:
Would opening an issue upstream be likely to get this fixed?

sjackman:
I've reported the bug upstream now. I guess we'll see.

sjackman:
Here's the bug reported upstream: http://lists.gnu.org/archive/html/bug-ncurses/2016-07/msg00001.html

sjackman:
Thomas E. Dickey wrote …

> hmm - the logs at 
> 
> https://gist.github.com/anonymous/3d5a6fa1fcb71d7d55f0afbabcc7b6b3#file-02-make-L2-L1440 
> 
> give the needed information. But my attempt to reproduce this with -j4 
> hasn't succeeded (I don't have hardware that could do -j20).

ThomasDickey:
There've been occasional problems with make -j{something}, which we've worked around by adding a dependency.  I investigated this one today, but didn't have any luck reproducing it.  I tried on Debian (no problem seen) and with a CentOS 6.7 (which it seems performs less well, but still "works").  Hints and patches are welcome.

sjackman:
@dunn This PR only deparallelizes `make install`. `make all` is still parallel. Merge?

dunn:
Go for it.

sjackman:
I don't have commit rights to dupes.

dunn:
Woops.  Merged as b1596c272e543ec68122b8b3c3219b4841e27930.

sjackman:
Thanks, Alex.

sjackman:
Egmont Koblinger wrote …
> Hi guys,
> 
> My 2 cents, not related to ncurses in particular:
> 
> 10+ years ago I used to work for a small Linux distribution. We built our build system from scratch. Each package was separately built in designated chroot (usually ./configure; make; make install – or whatever was required for that package), these kinds of builds were parallelized across multiple computers. It worked like a charm. We had a very few special cases (maybe 2-3 packages out of 1000) where the build was undeterministic for various weird reasons that we tracked down individually (the one I remember was that the 1 second precision of files' timestamps wasn't fine-grained enough, we needed to insert a "sleep 1.1" somewhere to make absolutely sure the timestamp increased).
> 
> At one point a colleague of mine insisted on trying "make -j2" (or -j4, can't recall – whatever) builds by default to speed them up. I firmly opposed, but we gave it a try. Out of the ~1000 packages we had we saw a failure after like perhaps the 10th package that we tried to build. That was the moment we reverted this change for good.
> 
> What's behind the story?
> 
> "make" (without -j) is deterministic and hence reproducible. Apart from a very few really tricky cases, if it works on the author's computer then it'll work on yours.
> 
> "make -j..." is undeterministic and hence unreproducible. Even for small projects the order of builds can vary in hundreds or thousands of ways. Noone is ever able to test them all. If it builds successfully on the author's machine many times, there's still absolutely no guarantee that it'll always build on yours. (Especially since the author probably doesn't use parallelization).
> 
> Getting Makefiles (or any other build rules) correct is tricky. There is absolutely no guarantee that these build rules are correct and solid across all kinds of parallelization and timings. And if something's not tested, you should not expect it to work correctly.
> 
> My concluision:
> 
> Feel free to use -j for your own project, provided that as soon as you face a glitch you immediately track it down, or even better, if you have some automated tools to generate the build rules that guarantee to always work.
> 
> Feel free to use -j for someone else's project if you know for sure that they also use it, or perhaps it's there by default in the build rules.
> 
> If you don't have a solid proof that the author uses -j or any other means that guarantee that it works: just forget it!
> 
> I'm not saying there's no such bug in ncurses. You've found it so apparently it's there. But I guess you could similarly find thousands of other projects that are broken similarly and it's just not feasible to get them all fixed (and make sure that they are indeed robustly fixed, no matter how the parallel build happens to go). Just accept that we live in an imperfect world, and stay safe by using serial builds unless you're absolutely certain you're safe to go parallel. Go and grab one more coffee or take a longer walk until the build completes, or use your remaining CPUs/cores to run a totally independent task. The amount you'd win with a successful parallel build is nothing compared to what you'd lose from uncaught/untraceable/unreproducible bugs of a failing parallel one.
> 
> Cheers,
> egmont
>

PR #635, opened 2016-07-20
nano: Depends on libmagic for Linuxbrew

sjackman:
Fix `nano: error while loading shared libraries: libmagic.so.1: cannot open shared object file`


sjackman:
Fix `nano: error while loading shared libraries: libmagic.so.1: cannot open shared object file`

DomT4:
Merged in https://github.com/Homebrew/homebrew-dupes/commit/36b6b7cd76a482319611eeb71e51f3134018a21c. Thanks Shaun!

sjackman:
Thanks for merging, Dominyk!

PR #640, opened 2016-08-04
openssh: add patch for ECDSA via PKCS#11

titanous:
This patch is required to use ECDSA keys over PKCS#11, for example with the Yubikey 4.

I've tested it and it works well, others have been using it with no issues for more than six months.

Upstream issue: https://bugzilla.mindrot.org/show_bug.cgi?id=2474


tdsmith:
Sorry; we don't accept feature patches. One way you could use Homebrew to install this software is to [maintain your own tap](https://github.com/Homebrew/brew/blob/master/share/doc/homebrew/How-to-Create-and-Maintain-a-Tap.md). The linked document has some advice; please let us know if you have any questions and we'll be happy to help you set things up.

PR #653, opened 2016-09-22
krb5 1.14.4

ericmlujan:
Updated krb5 to 1.14.4, which is the current release here at MIT.


apjanke:
Merged. Thank you for your contribution to Homebrew!

PR #663, opened 2016-10-10
file-formula: don't try to share libmagic

jesboat:
A previous commit (aa601b0) modified the formula to attempt to share the
library and database from the libmagic formula. Unfortunately, that doesn't
seem to work in all environments. In particular, with file 5.28 on el_capitan
in /opt/homebrew, file cannot be run:

    dyld: Library not loaded: /opt/homebrew/Cellar/file-formula/5.28/lib/libmagic.1.dylib
      Referenced from: /opt/homebrew/Cellar/file-formula/5.28/bin/file
      Reason: image not found

It doesn't look like there is out-of-the-box support for this in libmagic/file.
Even if there was, it might be tricky getting the versions right. I think
ideally we'd have a single brew which was part-keg-only, but I don't see a way
to do that either. I'm open to suggestions.

Test plan:

`brew install --build-from-source file-formula && brew test file-formula`

### DIFF
--- a/file-formula.rb
+++ b/file-formula.rb
@@ -8,6 +8,8 @@ class FileFormula < Formula
 
   head "https://github.com/file/file.git"
 
+  revision 1
+
   bottle do
     cellar :any
     sha256 "23e98e27a13e15e5f24c7c52fe7d6e0c49ec2cd53e572f4f2823a99af69eb593" => :el_capitan
@@ -17,14 +19,10 @@ class FileFormula < Formula
 
   keg_only :provided_by_osx
 
-  depends_on "libmagic"
-
   def install
     system "./configure", "--disable-dependency-tracking",
                           "--prefix=#{prefix}"
-    system "make", "install-exec"
-    system "make", "-C", "doc", "install-man1"
-    rm_r lib
+    system "make", "install"
   end
 
   test do


dunn:
It's working for me:
```
alex in sys27-mb at ~
🔥  /usr/local/opt/file-formula/bin/file ~/.bash_history
/Users/alex/.bash_history: ASCII text
```
What command(s) are failing for you?

dunn:
You might just need to reinstall libmagic.

jesboat:
`brew test` fails (as does any invocation of file.) Reinstalling libmagic was insufficient.

I think it works for you because you have Homebrew installed in /usr/local, and the configure (`m4/libtool.m4`, actually) blindly shoves /usr/local/lib into the search path. What does `otool -L /usr/local/opt/file-formula/bin/file` show? If it shows /usr/local/lib/libmagic, then I think it's effectively working by coincidence for you: the configure (unconditionally) stuck /usr/local/lib into the search path, and the libmagic there happens to be compatible because you chose to put homebrew in /usr/local/. If it shows /usr/local/Cellar/something or /usr/local/opt/libmagic/something then I don't know what's different between our setups.

dunn:
Even stranger, mine is also linking against the non-existed internal lib:
```
🌲  otool -L /usr/local/opt/file-formula/bin/file
/usr/local/opt/file-formula/bin/file:
        /usr/local/Cellar/file-formula/5.28/lib/libmagic.1.dylib (compatibility version 2.0.0, current version 2.0.0)
        /usr/lib/libz.1.dylib (compatibility version 1.0.0, current version 1.2.8)
        /usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1238.0.0)

ilovezfs:
I have fixed this in https://github.com/Homebrew/homebrew-dupes/pull/672. See the inline patch at the bottom if you're curious.

Issue #641, opened 2016-08-08
openssh 7.3p1: sshd won't take connections, @PREFIX@ in binary


philpennock:
With OpenSSH 7.3p1 installed from bottle, sshd will start and accept connections but kill all children; running `sudo /usr/local/sbin/sshd -D -d` we see:

```
debug1: Local version string SSH-2.0-OpenSSH_7.3
ssh_sandbox_child: sandbox_init: @PREFIX@/share/openssh/org.openssh.sshd.sb: profile not found [preauth]
```

That `@PREFIX@` comes from the patch `patch-sandbox-darwin.c-apple-sandbox-named-external.diff` and makes it through to the binary.

Making the `sshd` writable and running `hexedit` on it, to change the path to `/usr/local/etc/ssh/org.openssh.sshd.sb` without changing anything else, the bottled binary works and accepts connections without dying.

MikeMcQuaid:
We'll accept PRs for this but we're not actively working on it at this time.

DomT4:
Try `brew reinstall openssh`. I wasn't able to reproduce this locally, so maybe the bottle poured oddly for you locally or something.

philpennock:
A Bottle is a _binary_ package. It can't be "poured oddly" to change the contents of the binary (without being a major security issue).

I do see that I neglected to provide platform information: 10.10.5 Yosemite; I did not review all bottles to determine if this flaw only affected some, as I (naively?) assumed that the same build process would be used on all platforms, so that what's a bug on one is a bug on all.

I will put a PR on my backlog burner of tasks, thanks.

DomT4:
Sorry, I didn't realise you'd been maintaining Homebrew for quite some time now. Must have missed you in the team chat.

Homebrew performs [relocation on some strings](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/keg_relocate.rb) so that bottles are not always limited to the prefix they were built for. This has caused the occasional issue [before](https://github.com/Homebrew/legacy-homebrew/issues/41926). Forgetting temporarily that we use `@@HOMEBREW_PREFIX@@` as our marker rather than `@PREFIX@` I wondered if this was related.

As I said above, I cannot reproduce your issue. I look forwards to your PR.

milli:
I have also had to hack the binary to make this work on Yosemite (10.10.5).  Any chance there's a fix forth coming?

MikeMcQuaid:
> Any chance there's a fix forth coming?

[...]

PR #665, opened 2016-10-13
fix for https://github.com/Homebrew/homebrew-dupes/issues/641

mattmartini:
changed @PREFIX@/share/openssh/org.openssh.sshd.sb to
/System/Library/Sandbox/Profiles/org.openssh.sshd.sb
in:
patch-sandbox-darwin.c-apple-sandbox-named-external.diff
renamed to:
patch-sandbox-darwin.c-apple-sandbox-named-external-prefix-fix.diff

### DIFF
--- a/openssh.rb
+++ b/openssh.rb
@@ -28,8 +28,8 @@ class Openssh < Formula
   if OS.mac?
     # Both these patches are applied by Apple.
     patch do
-      url "https://raw.githubusercontent.com/Homebrew/patches/1860b0a74/openssh/patch-sandbox-darwin.c-apple-sandbox-named-external.diff"
-      sha256 "d886b98f99fd27e3157b02b5b57f3fb49f43fd33806195970d4567f12be66e71"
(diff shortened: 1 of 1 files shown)


dunn:
Sorry this fell off everyone's radar.  What is the difference between the two patches?

mattmartini:
I submitted a working patch (for Yosemite) (see prev comment) but there is an issue because the patch needs to be different for Sierra (text in patch needs to be "com.openssh.sshd.sb" instead of "org.openssh.sshd.sb") and I am not sure how to make the patch conditional on the OS version.

dunn:
You should be able to just put the patch(es) within conditional blocks like `if MacOS.version < :sierra`

MikeMcQuaid:
This formula is moving to homebrew/core in the next 24h, please reopen this PR there, sorry.

I'd suggest using `inreplace` in `install` and using the existing Apple patch if you want to modify based on the OS.

PR #666, opened 2016-10-14
Compile OpenLDAP with same overlays as Apple by default

lildude:
Installing the OpenLDAP formula doesn't currently give you a build that is identical or even close to the binary as shipped by Apple, particularly around the overlays, as we can see from:

```
$ /usr/libexec/slapd -VVV
@(#) $OpenLDAP: slapd 2.4.28 (Jul 30 2016 16:22:11) $
	root@osx208.apple.com:/Library/Caches/com.apple.xbs/Binaries/OpenLDAP/OpenLDAP-523~204/TempContent/Objects/servers/slapd

Included static overlays:
    accesslog
    dynid       (Unique to Apple)
    odlocales   (Unique to Apple)
    odusers     (Unique to Apple)
    auditlog
    constraint
    dds
    deref
    dyngroup
    nestedgroup (Unique to Apple)
    dynlist
    memberof
    ppolicy
    pcache
    refint
    retcode
    seqmod
    syncprov
    translucent
    unique
    valsort
Included static backends:
    config
    ldif
    monitor
    bdb
    hdb
    relay
$
```

... versus the homebrew/dupes/openldap install without any additional options:

```
$ brew install homebrew/dupes/openldap
[... truncated for brevity ...]
$ /usr/local/opt/openldap/libexec/slapd -VVV
@(#) $OpenLDAP: slapd 2.4.43 (Oct 11 2016 10:43:28) $
	lildude@[redacted]:/private/tmp/openldap-20161011-22180-o54lyh/openldap-2.4.43/servers/slapd

Included static overlays:
    syncprov
Included static backends:
    config
    ldif
    monitor
    mdb
    relay
$
```

Passing in all the allowed options doesn't get us much closer.

```
$ brew install homebrew/dupes/openldap --with-berkeley-db4 --with-memberof --with-unique --with-sssvlv
[... truncated for brevity ...]
$ /usr/local/opt/openldap/libexec/slapd -VVV
@(#) $OpenLDAP: slapd 2.4.43 (Oct 11 2016 10:50:11) $
	lildude@[redacted]:/private/tmp/openldap-20161011-87228-1h683k9/openldap-2.4.43/servers/slapd

Included static overlays:
    memberof
    sssvlv
    syncprov
    unique
Included static backends:
    config
    ldif
    monitor
    bdb
    hdb
    mdb
    relay
$
```

... and the other modules are not built as dynamically loadable modules so they can't be dynamically included either.

This difference means people can't take the configuration they use with the Apple build and use it with the Homebrew build, assuming they're not taking advantage of any of the Apple-specific overlays.

As the purpose of homebrew-dupes is to provide more up-to-date versions of the software Apple supplies, I think it makes sense for the OpenLDAP formula to get as close to the Apple build as possible by including the same overlays as the Apple build by default.

This PR does exactly that and ensures the resulting OpenLDAP is compiled with the same options as the Apple supplied version, with the exception of the `dynid`, `odlocales`, `odusers` and `nestedgroup` overlays as these are specific to the [Apple source](http://opensource.apple.com/source/OpenLDAP/) and haven't been included in the upstream source yet.

Instead of enabling the overlays, users can exclude overlays at installation using `--without-[overlay]` which is the equivalent of passing the `--enable-[overlay]=no` options to `./configure`.

I've not changed the backends.

This updated formula now produces:

```
$ brew install openldap.rb
[... truncated for brevity ...]
$ /usr/local/opt/openldap/libexec/slapd -VVV
@(#) $OpenLDAP: slapd 2.4.43 (Oct 14 2016 17:39:18) $
	lildude@[redacted]:/private/tmp/openldap-20161014-5156-1siyadm/openldap-2.4.43/servers/slapd

Included static overlays:
    accesslog
    auditlog
    constraint
    dds
    deref
    dyngroup
    dynlist
    memberof
    ppolicy
    pcache
    refint
    retcode
    seqmod
    syncprov
    translucent
    unique
    valsort
Included static backends:
    config
    ldif
    monitor
    mdb
    relay
```

... and this if you include the Berkley DB backend:

```
$ brew install openldap.rb --with-berkeley-db4
[... truncated for brevity ...]
$ /usr/local/opt/openldap/libexec/slapd -VVV
@(#) $OpenLDAP: slapd 2.4.43 (Oct 14 2016 17:43:31) $
	lildude@[redacted]:/private/tmp/openldap-20161014-68205-ha8tea/openldap-2.4.43/servers/slapd

Included static overlays:
    accesslog
    auditlog
    constraint
    dds
    deref
    dyngroup
    dynlist
    memberof
    ppolicy
    pcache
    refint
    retcode
    seqmod
    syncprov
    translucent
    unique
    valsort
Included static backends:
    config
    ldif
    monitor
    bdb
    hdb
    mdb
    relay
$
```

If someone was using the Apple build before, without taking advantage of any of the Apple-specific overlays, they should find their configuration will now work with this build without modification.

All testing has been performed on macOS Sierra 10.12.

/cc @mikemcquaid


openldap.rb
```diff
@@ -30,9 +44,15 @@ def install
       --localstatedir=#{var}
     ]
 
+    %w[accesslog auditlog constraint dds deref dyngroup dynlist memberof ppolicy proxycache refint retcode seqmod translucent unique valsort].each do |overlay|
+      if build.without? "#{overlay}"
+        args << "--enable-#{overlay}=no"
+      else
+        args << "--enable-#{overlay}"
+      end
+    end
```
MikeMcQuaid: I know the verbosity may kill you but the `args << "--enable-accesslog" if build.with? "accesslog"` format below would be better to follow. Additionally, if there's no particularly good reason to disable any of these you can remove them as an option altogether.

Given you're adding a bunch of new options at the same time you may want to do a `brew audit --strict` and adjust the options to the new format (i.e. `--with-accesslog`)

MikeMcQuaid:
Thanks for the detailed write-up @lildude!

lildude:
@MikeMcQuaid I can't think of any reason why you'd want to disable any of these overlays so I've changed things so they're all enabled by default with no option of disabling.  I've left enabling sssvlv as optional as I know not everyone wants server-side sorting.

I've also added a test which runs `sbin/slappasswd` which itself is just a symlink to `libexec/slapd`.  I've gone with this because it confirms `slapd` is operational without having to get into the realms of creating custom configuration just for the test or forcing the installation with Berkeley DB, which the default configuration files expect.

MikeMcQuaid:
Thanks @lildude! I changed your commit message to be more like our usual format: `openldap: compile with Apple default overlays.` Enjoy the weekend!

PR #667, opened 2016-10-16
tcl-tk: fix missing pkg-config build dependency.

kenh:
When configuring tcl-tk --with-x11, pkg-config should be listed as a
dependency so the configure script can find Xft2 and fontconfig.


dunn:
@BrewTestBot test this please

fxcoudert:
As far as I can see, `tcl-tk --with-x11` builds fine without `pkg-config`

kenh:
>As far as I can see, `tcl-tk --with-x11` builds fine without `pkg-config`

Forgive me if I didn't explain in greater detail, but the available
documentation emphasized brevity in pull-up requests.

Yes, tcl-tk --with-x11 builds file.  The configure script WANTS to use
Xft2 and fontconfig if they are available, but it doesn't need to.  But
even if they are available, the configure script won't make use of them
because pkg-config isn't listed as a build dependency.

--Ken

fxcoudert:
In my Sierra build a few minutes ago, I didn't see any warning about fontconfig or Xft2: it is not even looking for them I think (unless I missed something).

kenh:
>In my Sierra build a few minutes ago, I didn't see any warning about fontconfig or Xft2: it is not even looking for them I think (unless I missed something).

You can see it in the configure output when Tk is built.  I use the name
"Xft2" because that's what everyone I'm familiar with calls it, but
in the configure output it says:

checking whether to use xft... no

And if you dig into the Autoconf test for that, you'll see it calls
pkg-config.

--Ken

fxcoudert:
@BrewTestBot test this please

fxcoudert:
@BrewTestBot test this please

PR #680, opened 2016-11-05
nano: add without-nls option.

arabesc:


dunn:
Should this option remove the hard dependency on gettext?

bfontaine:
Ping on Alex’s comment above?

arabesc:
It's ready.

nano.rb
```diff
@@ -17,8 +17,10 @@ class Nano < Formula
     depends_on "autoconf" => :build
   end
 
+  option "without-nls", "Build without Native Language Support"
+
   depends_on "pkg-config" => :build
-  depends_on "gettext"
+  depends_on "gettext" if build.with? "nls"
```
ilovezfs:
```
depends_on "gettext" => :recommended
```

nano.rb
```diff
@@ -17,8 +17,10 @@ class Nano < Formula
     depends_on "autoconf" => :build
   end
 
+  option "without-nls", "Build without Native Language Support"
```
ilovezfs:
```
option "without-gettext", "Build without Native Language Support"
```

nano.rb
arabesc: Why is it better than `if build.with? "nls"`?

ilovezfs: because it's visible to the dependency system as it's bound to a formula

nano.rb
arabesc: Do you suggest to replace "without-nls" to "without-gettext" or to add a new option?

arabesc: I think that "without-nls" is more widespread and obvious.

ilovezfs: It has to match the `=> :recommended`, which creates the option. This just overrides the default description.

arabesc: I see. I have no expertise in the area to decide, so I suggest someone clone the changes and do it further in a right way.

ilovezfs: @arabesc I've updated the PR including fixing the head spec.

MikeMcQuaid:
This formula has been moved to homebrew/core. If @ilovezfs is cool on accepting it here it'll need to be moved over. Sorry!

PR #683, opened 2016-12-01
expect: Depends on tcl-tk for Linuxbrew

sjackman:
Fix the error:
```
checking for Tcl configuration... configure: error:
/usr/lib directory doesn't contain tclConfig.sh
```


MikeMcQuaid:

PR #684, opened 2016-12-01
lsof: Fix the location of tar for Linuxbrew

sjackman:
Fix `Error: Failure while executing: /usr/bin/tar xf lsof_4.89_src.tar`


MikeMcQuaid:

PR #685, opened 2016-12-01
screen: Depends on ncurses for Linuxbrew

sjackman:
Depends on ncurses for Linuxbrew
Fix `configure: error: no tgetent - no screen`

Include utmp.h in CFLAGS for Linuxbrew
Fix `error: dereferencing pointer to incomplete type 'struct utmp'`


MikeMcQuaid:

PR #686, opened 2016-12-01
heimdal: Depends on ncurses for Linuxbrew

sjackman:
Fix the error:
```
checking for tgetent in -lcurses... no
checking for tgetent in -lncurses... no
configure: error: libcurses or libncurses are required!
configure: error: ./configure failed for lib/libedit
```


sjackman:
```
Error: 1 problem in 1 formula
homebrew/dupes/heimdal:
  * Stable version URLs should not contain rc2
```
https://bot.brew.sh/job/Homebrew%20Dupes%20Pull%20Requests/367/version=el_capitan/testReport/junit/brew-test-bot/el_capitan/audit_homebrew_dupes_heimdal/

MikeMcQuaid:

PR #687, opened 2016-12-01
openldap: Depends on util-linux for Linuxbrew

sjackman:
Fix `slappasswd: error while loading shared libraries: libuuid.so.1`


MikeMcQuaid:

sjackman:
Great! That's excellent.

PR #691, opened 2016-12-15
Update to 2.7.2 version

mscalora:
Update to latest stable release version of nano


apjanke:
Superseded by #692, which also updated to 2.7.2. Sorry we missed this one! If you do a `brew update; brew upgrade`, you should pick up 2.7.2 now.

apjanke:
(I'd guess it was overlooked because the program name wasn't in the PR title, so it wouldn't pop out when scanning open issues. For future reference, we use the format "nano 2.7.2" for both the PR title and commit comment.)

PR #695, opened 2016-12-19
openssh: use secure urls

vszakats:


vszakats:
Not sure why the error on Yosemite only for the `mirror` and `homepage`. `mirror` didn't change in this update and `homepage` does have a correct Let's Encrypt certificate, which _should_ work okay. Maybe something transient?

apjanke:
Yeah, that's odd; sounds like possibly a transient error to me too. I queued up a re-run on the test bot.

apjanke:
Failed again on Yosemite, but this time the fetch and install steps succeeded, and only the `audit` step failed.

```
failed: brew audit homebrew/dupes/openssh --online

Stacktrace

        Error: 1 problem in 1 formula
homebrew/dupes/openssh:
  * The homepage is not reachable (curl exit code 35)
```

That does sound like a transient error, but I'm hesitant to override it. Will try rerunning it again in a while and see if it clears up.

apjanke:
Failed the same way on Yosemite only again. Same problem with https://github.com/Homebrew/homebrew-core/pull/8036. Sounds like there's something going on with the test-bot box or Yosemite itself.

ilovezfs:
See my comment here: https://github.com/Homebrew/homebrew-core/pull/8036#issuecomment-268950425.

vszakats:
Thanks for the updates! If this is Yosemite-specific curl issue, I agree with you to ignore it.

vszakats:
Is there any intention to merge this one? (following the linked and merged PR with the same issue)

PR #724, opened 2017-02-15
Make libiconv compile on Linux systems

GHPS:
Inhibit applying the Mac specific patches


MikeMcQuaid:
This is a Homebrew tap rather than Linuxbrew so this should be instead addressed in a Linuxbrew/homebrew-dupes fork. Sorry!

PR #731, opened 2017-03-03
openldap: fix berkeley-db4 dependency name

miguelarauj1o:
```
homebrew/dupes/openldap:
  * Dependency 'berkeley-db4' was renamed; use new name 'berkeley-db@4'.
```


miguelarauj1o:
Stacktrace from Jenkins:

``` bash
Error Message
failed: brew audit homebrew/dupes/openldap --online
Stacktrace
Error: 1 problem in 1 formula
homebrew/dupes/openldap:
  * The URL http://www.openldap.org/software/ could use HTTPS rather than HTTP
```

PR #742, opened 2017-03-31
Update to the new release of nano 2.8

mscalora:
GNU nano 2.8.0 "Axat" makes it easier to move around in
softwrapped lines: the Up and Down keys now step from
visual row to visual row instead of jumping between
logical lines, and the Home and End keys now move to
the start and end of a row, and only when already
there, then to the start and end of the logical line.
Furthermore, the screen can now scroll per row instead
of always per logical line.  On an entirely different
front: nano now makes use of gnulib, to make it build
on more platforms.  In short: there were many internal
changes, not many user-visible ones (apart form the
new softwrap navigation).  The conversion to gnulib
was done by Mike Frysinger, the softwrap overhaul by
David Ramsey.


MikeMcQuaid:
This was moved and merged in Homebrew/core.

PR #744, opened 2017-04-02
ntp 4.2.8p10: new formula

johnhawkinson:
The Network Time Protocol (NTP) Distribution.

This is a little tricky since:

* It replaces a MacOS component that is thoroughly busted. I'm sure it's thoroughly busted under 10.10 and under the strong impression that hasn't changed under 10.11 or 10.12, but I still live my life in Yosemite so I'm not really in a position to test that. I've been running the stock ntpd on my Yosemite machine for years, never bothered to package it though.

* It requires disabling the existing ntpd, which doesn't quite work in the `brew services `regime. Hopefully the use of Caveats is appropriate here -- I would assume so since the presence of the plist file causes `brew services` directions to be appended to the Caveats. The result is 3 things the user has to do.

---

Also, I set `PATH_PERL`  to `/usr/local/bin/perl` because I want the installed perl scripts (`ntptrace`, `ntpq`) not to crash on my system when `PERL5LIB` is set. I have a suspicion there might be an objection to this (although I'm not exactly sure why), but if so, please let me know what a useful remedy is that doesn't break those scripts on my system.

Furthermore, the upstream alleges 3 mirrored repositories (github, bitkeeper, rsync), and they are a) not mirrored and b) not up to date. I'm not sure what's going on there, maybe security issues for recent vulnerabilities? The bitkeeper repos are the most up-to-date, but it doesn't seem worth figuring out how to make that work in the formula since it doesn't actually gain anything. (Also the github repo is "special" because the default branch is `stable` and you need to explicitly check out `master` and I'm not sure what the right syntax for that is in homebrew's world.)

### DIFF
--- a/ntp.rb
+++ b/ntp.rb
@@ -1,0 +1,117 @@
+class Ntp < Formula
+  desc "The Network Time Protocol (NTP) Distribution"
+  homepage "https://www.eecis.udel.edu/~mills/ntp/html/"
+  # Arguments could be made for:
+  # http://ntp.org/documentation.html (has more unofficial stuff)
+  # http://ntp.org/  (more the protocol than the distribution)
+  url "http://www.eecis.udel.edu/~ntp/ntp_spool/ntp4/ntp-4.2/ntp-4.2.8p10.tar.gz"
+  version "4.2.8p10"
+  sha256 "ddd2366e64219b9efa0f7438e06800d0db394ac5c88e13c17b70d0dcdf99b99f"
+
+  devel do
+    # http://support.ntp.org/bin/view/Main/SoftwareDevelopment
+    url "http://www.eecis.udel.edu/~ntp/ntp_spool/ntp4/ntp-dev/ntp-dev-4.3.93.tar.gz"
+    sha256 "a07e73d7a3ff139bba33ee4b1110d5f3f4567465505d6317c9b50eefb9720c42"
+  end
+
+  head do
+    ## The git repo is broken/ancient, per https://github.com/ntp-project/ntp/issues/17
+    ## but the others aren't much better.
+    #
+    ## url "https://github.com/ntp-project/ntp.git"
+    ## Default stable branch: 1a399a03e674da08cfce2cdb847bfb65d65df237, Jan 24 2016
+    ## Non-default "master" branch: 9c75327c3796ff59ac648478cd4da8b205bceb77 Sun Jan 24 12:06:39 2016 +0000
+    #
+    ## bk://bk.ntp.org/ntp-stable
+    ## (4.2.8p10-win-beta1) 2017/03/21 Released by Harlan Stenn <stenn@ntp.org>
+    #
+    ## bk://bk.ntp.org/ntp-dev
+    ## (4.3.93) 2016/06/02 Released by Harlan Stenn <stenn@ntp.org>
+    #
+    ## rsync archive.ntp.org::ntp-dev-src
+    ## (4.3.92) 2016/04/27 Released by Harlan Stenn <stenn@ntp.org>
+
+    # Unsupported syntax?: depends_on cask: "bitkeeper"
+
+    depends_on "autoconf" => :build
+    depends_on "automake" => :build
+    depends_on "libtool" => :build
+    depends_on "lynx" => :build
+  end
+
+  depends_on "openssl"
+
+  def install
+    # Ensure homebrew perl is in #! lines of installed scripts.
+    ENV["PATH_PERL"] = "/usr/local/bin/perl"
+
+    system "./bootstrap" if build.head?
+
+    system "./configure", "--disable-debug",
+                          "--disable-dependency-tracking",
+                          "--disable-silent-rules",
+                          "--prefix=#{prefix}",
+                          "--with-openssl-libdir=#{Formula["openssl"].lib}",
+                          "--with-openssl-incdir=#{Formula["openssl"].include}"
+
+    system "make", "install"
+  end
+
+  def caveats
+    s = <<-EOS.undent
+    By default, OS X ships a broken version of ntpd that by design
+    does not control the system clock, but maintains the network
+    protocol and is intended to write a drift file (which it fails to
+    do under OS X 10.10). The drift file is read by pacemaker(8),
+    which is supposed to call adjtime() in a battery-efficient fashion
+    for laptops.
+
+    Installing this formula implies disabling the system ntpd as well as pacemaker:
+
+    sudo launchctl disable system/com.apple.pacemaker
+    sudo launchctl disable system/org.ntp.ntpd
+    EOS
+    s
+  end
+
+  plist_options :startup => true
+
+  # OS X uses /usr/libexec/ntp-wrapper to wait for appropriate network conditions and then
+  # invoke ntpd with various args. It's not clear that we need to resort to that. It also
+  # runs sntp first, which is definitely unnecessary given ntpd -g.
+  def plist; <<-EOS.undent
+    <?xml version="1.0" encoding="UTF-8"?>
+    <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+    <plist version="1.0">
+    <dict>
+      <key>KeepAlive</key>
+      <true/>
+      <key>Label</key>
+      <string>#{plist_name}</string>
+      <key>ProgramArguments</key>
+      <array>
+        <string>#{opt_sbin}/ntpd</string>
+        <string>-c</string>
+        <string>/private/etc/ntp-restrict.conf</string>
+        <string>-n</string>
+        <string>-g</string>
+        <string>-p</string>
+        <string>/var/run/ntpd.pid</string>
+        <string>-f</string>
+        <string>/var/db/ntp.drift</string>
+        <string>-l</string>
+        <string>/var/log/ntp.log</string>
+      </array>
+      <key>RunAtLoad</key>
+      <true/>
+    </dict>
+    </plist>
+    EOS
+  end
+
+  test do
+    # It's hard to do anything without network or clock-invasive stuff.
+    # Make sure we can print a usage message.
+    assert_match "usage: ", shell_output("#{sbin}/ntpdate -\? 2>&1", 2)
+  end
+end
--- a/ntp.rb
+++ b/ntp.rb
@@ -39,7 +39,10 @@ class Ntp < Formula
     depends_on "lynx" => :build
   end
 
+  option "with-snmp", "Build SNMP support"
+
   depends_on "openssl"
+  depends_on "net-snmp" if build.with? "snmp"
 
   def install
     # Ensure homebrew perl is in #! lines of installed scripts.
@@ -48,12 +51,17 @@ class Ntp < Formula
 
     system "./bootstrap" if build.head?
 
-    system "./configure", "--disable-debug",
-                          "--disable-dependency-tracking",
-                          "--disable-silent-rules",
-                          "--prefix=#{prefix}",
-                          "--with-openssl-libdir=#{Formula["openssl"].lib}",
-                          "--with-openssl-incdir=#{Formula["openssl"].include}"
+    args = [
+      "--disable-debug",
+      "--disable-dependency-tracking",
+      "--disable-silent-rules",
+      "--prefix=#{prefix}",
+      "--with-openssl-libdir=#{Formula["openssl"].lib}",
+      "--with-openssl-incdir=#{Formula["openssl"].include}",
+    ]
+    args << (build.with?("snmp") ? "-with-net-snmp-config" : "-with-net-snmp-config=no")
+
+    system "./configure", *args
 
     system "make", "install"
   end
--- a/ntp.rb
+++ b/ntp.rb
@@ -69,8 +69,8 @@ class Ntp < Formula
 
     Installing this formula implies disabling the system ntpd as well as pacemaker:
 
-    sudo launchctl disable system/com.apple.pacemaker
-    sudo launchctl disable system/org.ntp.ntpd
+      sudo launchctl disable system/com.apple.pacemaker
+      sudo launchctl disable system/org.ntp.ntpd
     EOS
     s
   end
--- a/ntp.rb
+++ b/ntp.rb
@@ -4,13 +4,13 @@ class Ntp < Formula
   # Arguments could be made for:
   # http://ntp.org/documentation.html (has more unofficial stuff)
   # http://ntp.org/  (more the protocol than the distribution)
-  url "http://www.eecis.udel.edu/~ntp/ntp_spool/ntp4/ntp-4.2/ntp-4.2.8p10.tar.gz"
+  url "https://www.eecis.udel.edu/~ntp/ntp_spool/ntp4/ntp-4.2/ntp-4.2.8p10.tar.gz"
   version "4.2.8p10"
   sha256 "ddd2366e64219b9efa0f7438e06800d0db394ac5c88e13c17b70d0dcdf99b99f"
 
   devel do
     # http://support.ntp.org/bin/view/Main/SoftwareDevelopment
-    url "http://www.eecis.udel.edu/~ntp/ntp_spool/ntp4/ntp-dev/ntp-dev-4.3.93.tar.gz"
+    url "https://www.eecis.udel.edu/~ntp/ntp_spool/ntp4/ntp-dev/ntp-dev-4.3.93.tar.gz"
     sha256 "a07e73d7a3ff139bba33ee4b1110d5f3f4567465505d6317c9b50eefb9720c42"
   end
 
--- a/ntp.rb
+++ b/ntp.rb
@@ -43,7 +43,8 @@ class Ntp < Formula
 
   def install
     # Ensure homebrew perl is in #! lines of installed scripts.
-    ENV["PATH_PERL"] = "/usr/local/bin/perl"
+    localperl="/usr/local/bin/perl"
+    ENV["PATH_PERL"] = localperl if File.exist? localperl
 
     system "./bootstrap" if build.head?
 


johnhawkinson:
> Also, I set PATH_PERL to /usr/local/bin/perl because I want the installed perl scripts (ntptrace, ntpq) not to crash on my system when PERL5LIB is set.

Well, that broke Jenkins.

```
Making install in ntpdc
clang -E -I. -I..  -I../include -I../lib/isc/include -I../lib/isc/nothreads/include -I../lib/isc/unix/include -I../sntp/libopts -I/usr/local/opt/openssl/include   ./nl_in.c | ./nl.pl > nl.c
/bin/sh: ./nl.pl: /usr/local/bin/perl: bad interpreter: No such file or directory
clang: error: unable to execute command: Broken pipe: 13
clang: error: clang frontend command failed due to signal (use -v to see invocation)
```
I did not realize PATH_PERL would override autoconf's check. Fixed in 1419b1b.
I had contemplated `"#{Formula["perl"].bin}/perl"` but decided that was worse than the disease.

> Error: 2 problems in 1 formula
homebrew/dupes/ntp:
>  * Stable: The URL http://www.eecis.udel.edu/~ntp/ntp_spool/ntp4/ntp-4.2/ntp-4.2.8p10.tar.gz could use HTTPS rather than HTTP
>  * Devel: The URL http://www.eecis.udel.edu/~ntp/ntp_spool/ntp4/ntp-dev/ntp-dev-4.3.93.tar.gz could use HTTPS rather than HTTP

Fixed in 53e5ed5. Why doesn't `brew audit --strict` catch this?

And one cosmetic fixup as well.

johnhawkinson:
Well. That is is "fun." Jenkins says the build fails on Sierra and El Cap., which I don't have easy access to, with:

```
libtool: link: clang -fno-strict-aliasing -g -Os -pipe -F/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.12.Internal.sdk/System/Library/PrivateFrameworks/ -F/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.12.Internal.sdk/System/Library/Frameworks/ -Wall -Wcast-align -Wcast-qual -Wmissing-prototypes -Wpointer-arith -Wshadow -Winit-self -Wstrict-overflow -D_P1003_1B_VISIBLE -Wstrict-prototypes -g -O2 -o ntpsnmpd netsnmp_daemonize.o ntpsnmpd.o ntpSnmpSubagentObject.o ntpsnmpd-opts.o  ../ntpq/libntpq.a ../libntp/libntp.a -lnetsnmpmibs -lnetsnmpagent -lnetsnmp -lcrypto.35 -lresolv -L/usr/local/opt/openssl/lib -lcrypto ../sntp/libopts/.libs/libopts.a
ld: library not found for -lcrypto.35
```

Which is because, I think, it is finding and linking against the system net-snmp installation which depends on the system OpenSSL:
```
checking for net-snmp-config... /usr/bin/net-snmp-config
checking for net-snmp-config path... /usr/bin/net-snmp-config
checking for net-snmp version... 5.6.2.1
```

Solution is to either build without snmp, or to depend on `net-snmp`. Sounds like an option to me...ae3ce6715c032da50409c99f45f54df5ce01af5d, done.

It seems like there should be a better way to detect this.
Googling the error messages was not a lot of help, nor searching github too helpful, although actually  https://github.com/Homebrew/homebrew-core/pull/4442 was indeed the issue, but it took me a bit to realize.

Maybe homebrew should shim in nonfunctional error-inducing versions of net-snmp-config and any other similar binaries that configure might test for in /usr/bin that link against difficult libraries like OpenSSL? Or...  I guess it's not frequent enough to merit an entry in the FAQ or Common Issues documents...

ntp.rb
```diff
@@ -0,0 +1,126 @@
...
+    does not control the system clock, but maintains the network
+    protocol and is intended to write a drift file (which it fails to
+    do under OS X 10.10). The drift file is read by pacemaker(8),
+    which is supposed to call adjtime() in a battery-efficient fashion
+    for laptops.
+
+    Installing this formula implies disabling the system ntpd as well as pacemaker:
+
+      sudo launchctl disable system/com.apple.pacemaker
+      sudo launchctl disable system/org.ntp.ntpd
```
MikeMcQuaid: There's other stuff worth commenting on but to avoid wasting your time: this, to me at least, makes this formula not suitable for inclusion in Homebrew. It's pretty unclear to me that disabling these will have no other adverse effects on the system either currently or in future so I'm a hard 👎 on recommending them.

johnhawkinson: OK. Where is this criterion stated, and what exactly is it? I thought the point of `homebrew-dupes` was to provide formulas for broken operating system software?

(MacPorts also has a Portfile: https://github.com/macports/macports-ports/blob/master/sysutils/ntp/Portfile
That makes me even more want to understand the philosophy. To put it differently, "If Homebrew won't package software I want, why am I using Homebrew instead of MacPorts again?")

---

On the merits, if we were to get to them, I suppose I'm not sure what the "no adverse effects" standard means here?

 `pacemaker(8)` is extremely clear about what the pacemaker program does, and I kind of presume we all understand what the `ntpd` program does (`pacemaker(8)` also discusses Apple's modifications in a limited way).

Is the question whether the existing `/System/Library/LaunchDaemons/com.apple.pacemaker.plist` and `/System/Library/LaunchDaemons/org.ntp.ntpd.plist` files do more than run `pacemaker` and `ntpd`? If so, that's pretty easy to show.

As for future effects on the system, whether adverse or not, it's pretty hard to predict the future. So I would wonder again what the standard is.

Thanks.

MikeMcQuaid: > Where is this criterion stated

In this comment. Our documentation cannot preemptively answer rare questions.

> and what exactly is it?

We shouldn't recommend/require people disable Apple-provided system services.

> MacPorts also has a Portfile: https://github.com/macports/macports-ports/blob/master/sysutils/ntp/Portfile

They don't recommend anything be disabled.

> To put it differently, "If Homebrew won't package software I want, why am I using Homebrew instead of MacPorts again?"

Ask yourself whether these questions make us more or less likely to help you. You can use this software as-is if you maintain you're own tap. You're asking us to maintain and support this software in an official tap.

> As for future effects on the system, whether adverse or not, it's pretty hard to predict the future.

I agree which is why I'd like to be conservative when it comes to system services.

>  I thought the point of homebrew-dupes was to provide formulas for broken operating system software?

Homebrew Dupes will be shortly moved into Homebrew Core.

johnhawkinson: I think I am "not getting it." I hope it's not futile to ask further:
>  Our documentation cannot preemptively answer rare questions.

Sure. One would like the criteria to be clear enough to not waste the time of contributors producing formulas that aren't going to be accepted. Since I suspect few people would find my own private tap, I would not have spent the time making the formula work if I'd known it didn't meet the criteria.

> We shouldn't recommend/require people disable Apple-provided system services.
> [MacPorts] don't recommend anything be disabled.

Err…what? It's OK to install a new ntpd as long as you don't recommend disabling the system ntpd? That seems like hiding from the facts we know -- is the formula somehow more acceptable without a caveat telling you how to use it? Similarly, MacPorts doesn't _explicitly_ recommend disabling the system ntpd, but if you try to run ntpd without stopping the system ntpd, you get an error:

```
 1 Apr 23:58:06 ntpd[72237]: ntpd 4.2.8p10@1.3728-o Sun Apr  2 01:00:39 UTC 2017 (1): Starting
 1 Apr 23:58:06 ntpd[72237]: Command line: /usr/local/opt/ntp/sbin/ntpd -c /private/etc/ntp-restrict.conf -g -p /var/run/ntpd.pid -f /var/db/ntp.drift -l /var/log/ntp.log
 1 Apr 23:58:06 ntpd[72238]: proto: precision = 1.000 usec (-20)
 1 Apr 23:58:06 ntpd[72238]: proto: fuzz beneath 0.077 usec
 1 Apr 23:58:06 ntpd[72238]: unable to bind to wildcard address :: - another process may be running - EXITING
```

So really it's a pretty clear implicit recommendation.

> I agree which is why I'd like to be conservative when it comes to system services.

It seems like a standard that can never be met. The result is a less useful package system.

Thanks.

ilovezfs:
For reference, here is how MacPorts packages ntp: https://github.com/macports/macports-ports/blob/master/sysutils/ntp/Portfile

MikeMcQuaid:
To quote myself earlier: 
> Ask yourself whether these questions make us more or less likely to help you. You can use this software as-is if you maintain you're own tap. You're asking us to maintain and support this software in an official tap.

To quote @johnhawkinson:

> To put it differently, "If Homebrew won't package software I want, why am I using Homebrew instead of MacPorts again?"
>  I hope it's not futile to ask further
> Err…what?
> It seems like a standard that can never be met. The result is a less useful package system.

I have no interest continuing this conversation with someone who speaks to me in this way and is irritating me and other maintainers.

--

I'm happy to package `ntp` with the caveats removed or worked around so they aren't needed. I'm happy to consider discussion on this PR if discussion can become more polite.

johnhawkinson:
> I'm happy to consider discussion on this PR if discussion can become more polite.

I apologize for offending you, @MikeMcQuaid (and others), and I did not intend my words to cross the bounds of politeness. Thank you for the offer to resume the discussion.

It's extremely frustrating to put time into something and be told that the work isn't valued, and especially so when the reason given just results in more questions that don't seem to have obvious answers to people who do not live and breath Homebrew day-in and day-out.
(I suppose it's my own fault, but often when I come to working on Homebrew, it is also from a point of frustration -- it seems like hours get spent tracing out the way things work with little discursions along the way (doc bug on `brew install -i`, error parsing in `brew services`, etc.) that take time don't produce a lot of result. I try to do the right thing by filing issues and pull requests, but it takes a toll. Similarly, I had thought about perhaps filing an issue or even a pull-request for the `net-snmp-config`/`libcrypto.35` problem, but now I just feel…discouraged. And as a non-Ruby programmer, working with Ruby often introduces small little challenges. Again, my own fault. And surely some of that frustration comes out in my language.)

But I'm not sure how else to ask some of these questions without being direct about it. I apologize if my directness is perceived as rudeness. It was my intent to value others time by asking questions directly,  rather than trying to construct them circuitously. I think I also don't quite understand the issues that are sensitive (should MacPorts never be mentioned?), so I'm not sure what to avoid.

For instance, the proposed package recommends disabling the system ntpd. MacPorts also packages ntpd and does not recommend disabling the system ntpd. You've said that it would be OK to package the software without recommending disabling the system ntpd. But anyone who tries to install and use the package will immediately run into the problem of a conflict with the system ntpd, and there should be some way to give them guidance as to a resolution. What is the right way to give them that guidance, please? Or to as the question differently, what's the reason to not recommend disabling a system package, as long as there is sufficient explanation about the reasoning to allow users to make an informed decision?

I feel like if I blindly move forward without understanding the philosophy — the reason Homebrew maintainers are uncomfortable with this packaging — then it will just lead to rejection. But also that if I ask the reason, my questions are perceived as excessively confrontational. So…CATCH-22?

_(I kind of feel like the kid who asks why the sky is blue and after a long series of questions we get to Rayleigh Scattering. Am I supposed to take that as a hint to stop asking further questions and give up my future career in atmospheric physics, or is it an invitation to keep asking? Words on a screen are not the best medium for conveying emotion. I wasn't offended, but clearly you were. It's not easy for me to see why.)_

Thanks again.

---

> I'm happy to package `ntp` with the caveats removed or worked around so they aren't needed

What would a workaround look like, please? Glancing at other packages in `homebrew-dupes`, I don't really see any similar cases. So it's hard to do my own research.

> For reference, here is how MacPorts packages ntp

@ilovezfs: Yeah, I mentioned that above in https://github.com/Homebrew/homebrew-dupes/pull/744#discussion_r109307384. Is there something in particular you wanted me to see to see in the Portfile that is helpful? I did review it, then and now, and I'm not seeing any differences that seem important.

Thank you.

MikeMcQuaid:
> I apologize for offending you, @MikeMcQuaid (and others), and I did not intend my words to cross the bounds of politeness. Thank you for the offer to resume the discussion.

I appreciate the apology and I also apologise for being less helpful than I could have been.

> It's extremely frustrating to put time into something and be told that the work isn't valued

That may have felt implied but it was not intended. There's just stuff that's not a good fit for Homebrew even although the formula is perfect because some users will blindly follow instructions and shoot themselves in the foot.

> But I'm not sure how else to ask some of these questions without being direct about it.

The directness is not a problem but it's side-swipes at Homebrew as a package manager, things like "err…what?". I get you were frustrated but that's not adding additional information to the issue beyond your frustration.

> what's the reason to not recommend disabling a system package, as long as there is sufficient explanation about the reasoning to allow users to make an informed decision?

I don't feel I could make an informed decision as I don't have enough insights into macOS to know that I could definitely disable those two services with no other negative consequences to my system at present or in future. Without being too arrogant, if I don't feel I can I don't think most of our users can either so they are likely to run commands somewhat blindly which may cause later issues. We try our best to be pretty strict about avoiding ever recommending fiddle with system-provided files and services.

> What would a workaround look like, please? Glancing at other packages in homebrew-dupes, I don't really see any similar cases. So it's hard to do my own research.

I realise this is probably unworkable for `ntp` but: running on an alternate port, for example.

Reopening this not because I'm more convinced it will be accepted but just to continue discussion until we decide either way on this. Thanks again for the apology, it means a lot ❤️

johnhawkinson:
> I realise this is probably unworkable for `ntp` but: running on an alternate port, for example.

Yes, that is unworkable in any practical sense.

Another approach which I find troublesome is to direct the user to go to `System Preferences` > `Date & Time` and disable the `Set date and time automatically` checkbox. Doing so disables Apple's NTP and pacemaker as well.

I find it troublesome for both rational and technical reasons:
* Date and Time are still being set automatically, yet the checkbox is inactive. To me this makes debugging confusing, and I could imagine a case where a third person is looking at a problem with a machine and wants to ensure time synchronization is off (why?) and goes and checks the checkbox, but it turns out that ntp is running seperately.
* I believe (not 100%, and definitely not sure beyond Yosemite) that `System Preferences` constructs an appropriate set of `/etc/ntp.conf` and related files; certainly it modifies them. If it's turned off, I'm not sure it's safe to depend on the existence of those configuration files.

But none of these are insurmountable. Perhaps that is more palatable to Homebrew?

Setting aside the "err...what?" language, the words after it try to ask an important question. If we can't give explicit direction to users about MacOS internals because we're not sufficiently confident, can we at least give guidance and explain issues (without explicit commands that users might blindly run)? The impression I get is that Homebrew would rather say nothing at all, but I find it very hard to understand why there is no place for informed guidance from the formula's author and users. Or am I just not finding the right place (if it is not Caveats)? Is it paragraph-length comments in the formula?

MikeMcQuaid:
> Perhaps that is more palatable to Homebrew?

That would be, yes.

> The impression I get is that Homebrew would rather say nothing at all, but I find it very hard to understand why there is no place for informed guidance from the formula's author and users. Or am I just not finding the right place (if it is not Caveats)? Is it paragraph-length comments in the formula?

There is no place because things that require a lot of explanation are generally not a good fit for a package manager's core repository (as mentioned: I think we may have found a workaround that means this is).

johnhawkinson:
So where does a discussion of unsetting the checkbox go. Caveats?

MikeMcQuaid:
> So where does a discussion of unsetting the checkbox go. Caveats?

Yup 👍

ntp.rb
```diff
@@ -0,0 +1,126 @@
+class Ntp < Formula
+  desc "The Network Time Protocol (NTP) Distribution"
+  homepage "https://www.eecis.udel.edu/~mills/ntp/html/"
+  # Arguments could be made for:
+  # http://ntp.org/documentation.html (has more unofficial stuff)
+  # http://ntp.org/  (more the protocol than the distribution)
```
MikeMcQuaid: can remove these comments.

ntp.rb
```diff
@@ -0,0 +1,126 @@
...
+  url "https://www.eecis.udel.edu/~ntp/ntp_spool/ntp4/ntp-4.2/ntp-4.2.8p10.tar.gz"
+  version "4.2.8p10"
+  sha256 "ddd2366e64219b9efa0f7438e06800d0db394ac5c88e13c17b70d0dcdf99b99f"
+
+  devel do
+    # http://support.ntp.org/bin/view/Main/SoftwareDevelopment
```
MikeMcQuaid: can remove this comment

ntp.rb
```diff
@@ -0,0 +1,126 @@
...
+    ## bk://bk.ntp.org/ntp-stable
+    ## (4.2.8p10-win-beta1) 2017/03/21 Released by Harlan Stenn <stenn@ntp.org>
+    #
+    ## bk://bk.ntp.org/ntp-dev
+    ## (4.3.93) 2016/06/02 Released by Harlan Stenn <stenn@ntp.org>
+    #
+    ## rsync archive.ntp.org::ntp-dev-src
+    ## (4.3.92) 2016/04/27 Released by Harlan Stenn <stenn@ntp.org>
+
+    # Unsupported syntax?: depends_on cask: "bitkeeper"
```
MikeMcQuaid: Can remove all these comments

ntp.rb
```diff
@@ -0,0 +1,126 @@
...
+    depends_on "autoconf" => :build
+    depends_on "automake" => :build
+    depends_on "libtool" => :build
+    depends_on "lynx" => :build
+  end
+
+  option "with-snmp", "Build SNMP support"
+
+  depends_on "openssl"
+  depends_on "net-snmp" if build.with? "snmp"
```
MikeMcQuaid: Use `depends_on "net-snmp" => :optional` and name the option `"with-net-snmp"`. This is annoying but works around quirks in our dependency resolution system.

ntp.rb
```diff
@@ -0,0 +1,126 @@
...
+
+  def install
+    # Ensure homebrew perl is in #! lines of installed scripts.
+    localperl="/usr/local/bin/perl"
+    ENV["PATH_PERL"] = localperl if File.exist? localperl
```
MikeMcQuaid: Let's remove this; it's fine to use the system Perl here.

ntp.rb
```diff
@@ -0,0 +1,126 @@
...
+
+    args = [
+      "--disable-debug",
+      "--disable-dependency-tracking",
+      "--disable-silent-rules",
+      "--prefix=#{prefix}",
+      "--with-openssl-libdir=#{Formula["openssl"].lib}",
+      "--with-openssl-incdir=#{Formula["openssl"].include}",
+    ]
+    args << (build.with?("snmp") ? "-with-net-snmp-config" : "-with-net-snmp-config=no")
```
MikeMcQuaid: Turn this into an `if`/`else` to make it easier to follow.

ntp.rb
```diff
@@ -0,0 +1,126 @@
...
+
+    system "./configure", *args
+
```
MikeMcQuaid: Can remove this empty line.

ntp.rb
```diff
@@ -0,0 +1,126 @@
...
+    which is supposed to call adjtime() in a battery-efficient fashion
+    for laptops.
+
+    Installing this formula implies disabling the system ntpd as well as pacemaker:
+
+      sudo launchctl disable system/com.apple.pacemaker
+      sudo launchctl disable system/org.ntp.ntpd
+    EOS
+    s
+  end
```
MikeMcQuaid: Let's just remove all the caveats (as requested by @ilovezfs)

ntp.rb
```diff
@@ -0,0 +1,126 @@
...
+        <string>/var/db/ntp.drift</string>
+        <string>-l</string>
+        <string>/var/log/ntp.log</string>
+      </array>
+      <key>RunAtLoad</key>
+      <true/>
+    </dict>
+    </plist>
+    EOS
+  end
```
MikeMcQuaid: Let's just remove the plist (as requested by @ilovezfs)

ntp.rb
```diff
@@ -0,0 +1,126 @@
...
+
+  test do
+    # It's hard to do anything without network or clock-invasive stuff.
+    # Make sure we can print a usage message.
```
MikeMcQuaid: Can remove these comments. Also, network access is fine; verifying connection to a remote, standard NTP server would be idea.

Issue #741, opened 2017-03-28
OpenSSH 7.5 --with-libressl still links against OpenSSL


jwm:
I've been using `--with-libressl` with the OpenSSH formula for some time now, but noticed that the formula fails to link against LibreSSL (and instead uses OpenSSL) with 7.5p1:

```
[jwm@hobbes:s007 ~> brew install openssh --with-libressl | cat -
Updating Homebrew...
==> Auto-updated Homebrew!
Updated 1 tap (caskroom/cask).
No changes to formulae.

==> Installing openssh from homebrew/dupes
==> Using the sandbox
==> Downloading https://ftp.openbsd.org/pub/OpenBSD/OpenSSH/portable/openssh-7.5p1.tar.gz
Already downloaded: /Users/jwm/Library/Caches/Homebrew/openssh-7.5p1.tar.gz
==> Downloading https://raw.githubusercontent.com/Homebrew/patches/1860b0a74/openssh/patch-sandbox-darwin.c-apple-sandbox-named-external.diff
Already downloaded: /Users/jwm/Library/Caches/Homebrew/openssh--patch-d886b98f99fd27e3157b02b5b57f3fb49f43fd33806195970d4567f12be66e71.diff
==> Downloading https://raw.githubusercontent.com/Homebrew/patches/d8b2d8c2/openssh/patch-sshd.c-apple-sandbox-named-external.diff
Already downloaded: /Users/jwm/Library/Caches/Homebrew/openssh--patch-3505c58bf1e584c8af92d916fe5f3f1899a6b15cc64a00ddece1dc0874b2f78f.diff
==> Patching
==> Applying patch-sandbox-darwin.c-apple-sandbox-named-external.diff
patching file sandbox-darwin.c
Hunk #1 succeeded at 63 (offset 1 line).
==> Applying patch-sshd.c-apple-sandbox-named-external.diff
patching file sshd.c
Hunk #1 succeeded at 619 (offset -94 lines).
==> ./configure --with-libedit --with-kerberos5 --prefix=/usr/local/Cellar/openssh/7.5p1 --sysconfdir=/usr/local/etc/ssh --with-pam --with-ssl-dir=/usr/local/opt/libressl
==> make
==> make install
🍺  /usr/local/Cellar/openssh/7.5p1: 32 files, 4.0MB, built in 1 minute 45 seconds
```

```
[jwm@hobbes:s007 ~> otool -L /usr/local/bin/ssh | grep -ie ssl
[...]

PR #746, opened 2017-04-02
Standardize on OpenSSL

jwm:


jwm:
This is a followup to #741.

MikeMcQuaid:
This is being done as part of https://github.com/Homebrew/homebrew-core/pull/11282 but thanks for the PR.

